PR #2, opened 2019-03-22
Chapter 4 exercises

gildedgardenia:
Chapter 4 Exercises are ready for review! There are 3 questions and one of those is a multi-part question.


src/chapters/data-and-variables/chapter-4-exercises.rst
```diff
@@ -0,0 +1,35 @@
...
+   * - Distance to the Moon (km)
+     - 384,400
+
+#. Assign a variable named spaceShuttleName the value of the name of the space shuttle.
+
+#. For each value in the table, if you were to assign them to variables in Javascript, what would be the data type?
+
+#. Time to calculate some important values for our space shuttle!
+
+   #. How long will it take to reach Mars?
```
welzie: More direction would probably be better here.
- How do we want them to calculate these?
- What are the formulas/algorithms to use?
- Do we need to give them more variables?

welzie:
I think the file should be named `exercises.rst` instead of `chapter-4-exercises.rst` so that the url is `/intro-to-professional-web-dev/docs/chapters/data-and-variables/exercises`

PR #3, opened 2019-03-25
Booleans and conditionals

chrisbay:
Chapter 5 is ready for review!


src/chapters/booleans-and-conditionals/booleans.rst
```diff
@@ -0,0 +1,209 @@
...
+
+Boolean Values
+--------------
+
+.. index:: ! true, ! false, ! boolean
+
+.. index::
+   single: boolean; value
+
+The JavaScript data type for storing true and false values is ``boolean``, named after the British mathematician, George Boole. George Boole created `Boolean Algebra <https://en.wikipedia.org/wiki/Boolean_algebra>`_ which is the basis of all modern computer arithmetic.
```
jimflores5: "...values is called ``boolean``..."

src/chapters/booleans-and-conditionals/booleans.rst
```diff
@@ -0,0 +1,209 @@
...
+
+      boolean
+      string
+
+Boolean Conversion
+------------------
+
+.. index:: ! Boolean(), type conversion
+
+As with the number and string data types, the boolean type has a type conversion function, ``Boolean()``. It works similarly to ``Number()`` and ``String()``, attempting to convert a non-boolean value to a boolean.
```
jimflores5: Phrasing suggestion: "As with the **number** and **string** data types, the **boolean** type also has a conversion function, ``Boolean()``."

src/chapters/booleans-and-conditionals/booleans.rst
```diff
@@ -0,0 +1,209 @@
...
+      console.log(Boolean("true"));
+      console.log(Boolean("TRUE"));
+      console.log(Boolean(0));
+      console.log(Boolean(1));
+      console.log(Boolean(''));
+      console.log(Boolean('LaunchCode'));
+
+   `Run this program at repl.it <https://repl.it/@launchcode/Boolean-Type-Conversion>`_
+
+   Under which conditions does ``Boolean()`` convert a string to ``true``?
```
jimflores5: This question seems to be a concept check, which should be outside of the 'Try It' box.

src/chapters/booleans-and-conditionals/booleans.rst
```diff
@@ -0,0 +1,209 @@
...
+
+.. index::
+   single: boolean; expression
+
+.. index::
+   single: operator; equality
+
+.. index:: ! ==
+
+A **boolean expression** is an expression that evaluates to a boolean value. The equality operator, ``==``, compares two values and produces a boolean value related to whether the two values are equal to one another.
```
jimflores5: I'll defer to Sally's comment for the first sentence.  ;-)

For the second sentence, how about, "The equality operator, ``==``, compares two values and returns ``true`` or ``false`` depending on whether the values are equal."

src/chapters/booleans-and-conditionals/booleans.rst
```diff
@@ -0,0 +1,209 @@
...
+   **Output**
+
+   ::
+
+      false
+
+.. index::
+   single: operator; comparison
+
+The ``==`` operator is one of six common **comparison operators**.
```
jimflores5: This could be emphasized via larger text or as a sub-heading to 5.1.3.

src/chapters/booleans-and-conditionals/booleans.rst
```diff
@@ -0,0 +1,209 @@
...
+
+.. index:: ==, ! !=, ! <, ! >, ! <=, ! >=
+
+.. list-table:: Comparison Operators
```
jimflores5: The table title "Comparison Operators" appears as light-grey on the webpage.  Could we darken the text to make it easier to see?  Alternatively, the table title seems redundant given the intro line, so the title could be removed without losing clarity.

src/chapters/booleans-and-conditionals/booleans.rst
```diff
@@ -0,0 +1,209 @@
...
+     - ``5 <= 7``
+
+       ``5 <= 5``
+
+       ``'a' <= 'b'``
+
+       ``'a' <= 'a'``
+
+
+Although these operations are probably familiar to you, the JavaScript symbols are different from the mathematical symbols. A common error is to use a single equal sign (``=``) instead of a double equal sign (``==``). Remember that ``=`` is an *assignment* operator and ``==`` is a *comparison* operator. Also note that ``=<`` and ``=>`` are not recognized operators.
```
jimflores5: Remove 'to you' from the phrase.

src/chapters/booleans-and-conditionals/equality.rst
```diff
@@ -0,0 +1,118 @@
...
+
+Equality
+========
+
+Loose Equality With ``==``
+--------------------------
+
+.. index:: equality, ==
+
+In the section :ref:`booleans`, we learned about the comparison operators ``==`` and ``!=``, which test whether two values are equal or not equal, respectively. We also tipped you off to some quirks of the ``==`` operator. Specifically, when its operands are of different types, then the returned boolean value may not be what you expect.
```
jimflores5: Grammar: 'We' in the first sentence implies a role of learners, but the 'we' in the second sentence implies a teacher role.  

To avoid this conflict, we (teachers) should remove the 'we' from the second sentence.  :-) 'We' could try something like, "However, there are some quirks with using the ``==`` operator."

The third sentence can return to the we (learners) role: "These often occur when we use ``==`` to compare different data types."

src/chapters/booleans-and-conditionals/equality.rst
```diff
@@ -0,0 +1,118 @@
...
+   ::
+
+      true
+      true
+      true
+
+.. index::
+   single: equality; loose
+
+When we give ``==`` operands of different data types, it will still try to make a comparison. In order to properly do so, the values must be both be of the same type. JavaScript will implicitely convert the operands so that they are of the same type, and then compare the converted values to generate ``true`` or ``false``. For this reason, the ``==`` operator is often said to measure **loose equality**.
```
jimflores5: Second sentence suggestion: "To do this, JavaScript will implicitly convert the operands into the *same data type* before making the comparison and returning ``true`` or ``false``."

src/chapters/booleans-and-conditionals/equality.rst
```diff
@@ -0,0 +1,118 @@
...
+   ::
+
+      true
+      true
+      false
+
+.. index::
+   single: equality; non-transitive
+
+This example demonstrates that ``==`` is **non-transitive**. We intuitively think of the concept of equality as being transitive. In other words, if A and B are equal, and B and C are equal, then A and C should also be equal. The example above demonstrates that this is **not** the case for the ``==`` operator.
```
jimflores5: I will defer to Sally's comments for this paragraph.

src/chapters/booleans-and-conditionals/equality.rst
```diff
@@ -0,0 +1,118 @@
...
+      true
+      true
+      false
+
+.. index::
+   single: equality; non-transitive
+
+This example demonstrates that ``==`` is **non-transitive**. We intuitively think of the concept of equality as being transitive. In other words, if A and B are equal, and B and C are equal, then A and C should also be equal. The example above demonstrates that this is **not** the case for the ``==`` operator. 
+
+Since ``==`` does not follow rules that we typically associate with quality, unexpected results can occur when relying on it in our programs. Thankfully, JavaScript provides another operator that returns more predictable results.
```
jimflores5: "equality" instead of "quality"

Suggestion: "...unexpected results may occur if ``==`` is used in a program."

src/chapters/booleans-and-conditionals/equality.rst
```diff
@@ -0,0 +1,118 @@
...
+.. index::
+   single: equality; strict
+
+For this reason, the ``===`` operator is often said to measure **strict equality**.
+
+.. index:: ! !==
+
+Just as equality operator ``==`` has the inequality operator ``!=``, there is also a strict inquality operator, ``!==``. The boolean expression ``a !== b`` returns ``true`` when the two operands are of different types, or if they are of the same type and have different values. 
+
+.. tip:: Use ``===`` and ``!==`` whenever possible. In this book we will use these strict operators over the loose operators from now on.
```
jimflores5: Opinion: Style the text in the tip box to ### REALLY EMPHASIZE the use of ``===`` and ``!==``.

src/chapters/booleans-and-conditionals/equality.rst
```diff
@@ -0,0 +1,118 @@
...
+   #. ``true``
+   #. ``false``
+   #. ``"true"``
+   #. ``"false"``
+
+.. admonition:: Question
+
+   What is the difference between ``==`` and ``===``?
+
+   #. There is no difference; they work exactly the same
```
jimflores5: There is no difference.  They work exactly the same.

src/chapters/booleans-and-conditionals/equality.rst
```diff
@@ -0,0 +1,118 @@
...
+   #. Only ``===`` throws an error of its arguments are of different types
```
jimflores5: "error if" instead of "error of".

src/chapters/booleans-and-conditionals/equality.rst
```diff
@@ -0,0 +1,118 @@
...
+   #. ``==`` will try to convert values of different types to be the same type, while ``===`` does no such conversion
```
jimflores5: ``==`` tries to convert different data types into the same type, while ``===`` does not.

src/chapters/booleans-and-conditionals/logical-operators.rst
```diff
@@ -0,0 +1,273 @@
...
+
+      true
+      false
+      false
+
+In line 1, ``7 > 5 && 5 > 3`` returns ``true`` because both ``7 > 5`` and ``5 > 3`` are ``true`` individually.
+
+In line 2, ``7 > 5 && 2 > 3`` returns ``false`` because one of the two expressions, ``2 > 3``, is ``false``.
+
+Like line 2, line 3 returns ``false`` because both sub-expressions are ``false``. Notice in this example that we can mix and match data types however we like, as long as both sides of the ``&&`` expression are themselves boolean expressions.
```
jimflores5: Phrasing: Remove "Like line 2".  Also, remove "in this example", since it implies that there may be a case where we CANNOT mix and match data types.

src/chapters/booleans-and-conditionals/logical-operators.rst
```diff
@@ -0,0 +1,273 @@
...
+
+::
+
+   true
+   true
+   false
+
+Lines 1 and 2 both return ``true`` because at least one of the comparison expressions joined by ``||`` is ``true``. Line 3 returns ``false`` because both sub-expressions are ``false``.
+
+.. warning:: The single symbols ``&`` and ``|`` are themselves valid JavaScript operators, so accidentally leaving off one of the symbols off when typing ``&&`` or ``||`` will not result in an error. The resulting value of the expression will be different than expected though.
```
jimflores5: Phrasing: Retain only one 'off' in the "...leaving" phrase.  Consider replacing 'themselves' with 'also'.

Phrasing: "However, the result will be different than expected."

Following this warning, a concrete example of using & vs. && or | vs. || would be helpful.  Alternatively, encourage the students to try for themselves on repl.it.

src/chapters/booleans-and-conditionals/logical-operators.rst
```diff
@@ -0,0 +1,273 @@
...
+
+Operator Precedence
+-------------------
+
+.. index::
+   single: operator; precedence
+
+We now have a number of operators in our toolkit. It is important to understand how these operators relate to each other with respect to **operator precedence**. Operator precedence is the set of rules that dictate in which order the operators are applied.
+
+JavaScript will always apply the logical NOT operator, ``!``, first. It then applies the arithmetic operators (``**`` is highest, then ``*`` and ``/``, then ``+`` and ``-``). Next comes the comparison operators. Finally, the logical AND and OR are applied last.
```
jimflores5: Phrasing: "JavaScript always applies the logical NOT operator, ``!``, first. Next, it applies the arithmetic operators, followed by the comparison operators. The logical AND (``&&``) and OR (``||``) are applied last."

There is no need to explain the arithmetic order of operations, since that information is show in the table below.

src/chapters/booleans-and-conditionals/logical-operators.rst
```diff
@@ -0,0 +1,273 @@
...
+   * - (lowest)
+     - Logical OR
+     - ``||``
+
+Truth Tables
+------------
+
+.. index:: ! truth table
+
+**Truth tables** help us understand how logical operators work by calculating all of the possible return values of a boolean expression. Let's look at the truth table for ``&&``, which displays assumes we have two boolean expressions, A and B, joined by ``&&``.
```
jimflores5: Typo: Remote 'displays'.  

"Let's look at the truth table for ``&&``, which assumes we have two boolean expressions, A and B, joined by ``&&``."

src/chapters/booleans-and-conditionals/logical-operators.rst
```diff
@@ -0,0 +1,273 @@
...
+        - ``false``
+
+Consider the first row of the truth table. This row states that if A is true and B is true, then A && B is true. This is a fact, regardless of what boolean expressions A and B might actually be. The two middle rows demonstrate that if either A or B is false, then A && B is false. (If this idea is hard to grasp, try substituting actual expressions for A and B.)
+
+Check Your Understanding
+------------------------
+
+.. admonition:: Question
+
+   Complete the truth table for ``||``.
```
jimflores5: Formatting: "Truth table for ``||``" is redundant after the instruction "Complete the truth table for ``||``."

src/chapters/booleans-and-conditionals/logical-operators.rst
```diff
@@ -0,0 +1,273 @@
...
+      * - ``false``
+        - ``true``
+        -
+      * - ``false``
+        - ``false``
+        -
+
+.. admonition:: Question
+
+   Which of the following properly expresses the precedence of operators (using parentheses) in the following expression?
```
jimflores5: Rephrasing: Replace "precedence of operators" with "order of operations", which is a more common term.

src/chapters/booleans-and-conditionals/booleans.rst
```diff
@@ -0,0 +1,209 @@
+.. _booleans:
+
+Booleans
+========
+
```
jimflores5: From the objectives: "Use ``Boolean`` to convert non-boolean values". 

Is there ever an explicit concept check for this idea?

src/chapters/booleans-and-conditionals/conditionals.rst
```diff
@@ -0,0 +1,233 @@
...
+
+.. index:: ! conditional
+
+A the beginning of this chapter, we motivated our exploration of booleans in JavaScript by observing that having a formal notion of truth and falsehood would allow us to write code that executes conditionally. In other words, we want to be able to write code that only executes when a given condition is ``true``.
+
+Again, here is our motivating example:
+
+.. admonition:: Example
+
+   Consider a banking application that can remind you when a bill is due. The application will notify the user that a bill is due soon, but *only if* the bill has not already been paid.
```
jimflores5: Earlier rephrasing suggestion: "Consider a banking application that reminds you to pay a bill on time. The app sends you a notification _only if_ you miss the due date."

src/chapters/booleans-and-conditionals/conditionals.rst
```diff
@@ -0,0 +1,233 @@
...
+
+We summarized the condition as follows: Send a notification of an upcoming bill if the statement "the bill is unpaid" is true.
+
+In such a program, booleans would allow us to write JavaScript that represents the conditional "the bill is unpaid", and based on the truth of such a statement, execute or skip code that notifies the user of an upcoming due date.
```
jimflores5: In such a program, JavaScript uses booleans to represent the conditional "the bill is unpaid".  Based on the truth of this statement, the program executes or skips the code for notifying the user.

src/chapters/booleans-and-conditionals/conditionals.rst
```diff
@@ -0,0 +1,233 @@
...
+The most basic form of a conditional is an **if statement**. Here's how to create one in JavaScript:
+
+.. figure:: figures/if.png
+   :height: 400px
+   :alt: The structure of a conditional with an if statement
+
+Let's look at each component of this new syntax.
+
+- The ``if`` statement consists of a header line and a body. The header line begins with the keyword ``if`` followed by a boolean expression enclosed in parentheses. 
+- ``condition`` is a boolean expression, that is, something that evaluates to either ``true`` or ``false``.
```
jimflores5: ``condition`` is a boolean expression that evaluates to either ``true`` or ``false``.

src/chapters/booleans-and-conditionals/conditionals.rst
```diff
@@ -0,0 +1,233 @@
...
+- The statements that follow the condition, within ``{}``, make up a **code block**. The code within the brackets will be executed when the condition evaluates to ``true``, and will be skipped otherwise.
```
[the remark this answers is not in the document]
jimflores5: "The code within the brackets {} will be executed when the condition evaluates to ``true``. When the condition evaluates to ``false``, the code within the brackets is ignored."

src/chapters/booleans-and-conditionals/conditionals.rst
```diff
@@ -0,0 +1,233 @@
...
+
+And here's an explicit example that mimics our banking example.
```
jimflores5: Here is an explicit example that mimics our banking program.

src/chapters/booleans-and-conditionals/conditionals.rst
```diff
@@ -0,0 +1,233 @@
...
+   .. sourcecode:: js
+
+      if (num % 2 === 0 && num > 3) {
+         console.log(num, "is even");
+         console.log(num, "is greater than 3");
+      }
+
+While not required, the code within a conditional code block is typically indented to make it more readable. Similarly, it is a common convention to place the opening ``{`` at the end of the first line, and the closing ``}`` on a line of its own following the last line of the code block. 
+
+You should follow such conventions, even though not doing so won't create an error. To see why, compare the readability of this example, which is functionally equivalent to the one above.
```
jimflores5: Double negative dance.  How about, "You should follow such conventions, even though ignoring them will not create an error."

src/chapters/booleans-and-conditionals/nested-conditionals.rst
```diff
@@ -0,0 +1,68 @@
+Nested Conditionals
+===================
+
+We can write code with more complex branching behavior by combining conditionals, and in particular by nesting conditionals. Let's see how this works by tackling the following problem.
```
jimflores5: Emphasize _nesting_.

src/chapters/booleans-and-conditionals/nested-conditionals.rst
```diff
@@ -0,0 +1,68 @@
...
+
+.. admonition:: Example
+
+    Write code that prints different messages based on the value of a number variable. If the number is odd, print nothing. If it is even, print "EVEN!" If it is also positive print "POSITIVE!".
```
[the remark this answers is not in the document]
jimflores5: Not sure about the double punctuation.  My vote is to let the !'s stand on their own.

src/chapters/booleans-and-conditionals/nested-conditionals.rst
```diff
@@ -0,0 +1,68 @@
...
+   if (num % 2 === 0) {
+       console.log("EVEN!");
+
+       if (num > 0) {
+           console.log("POSITIVE!");
+       }
+   }
+
+Try this with a bunch of values (even, odd, positive, negative) to ensure it works as desired. Nice, huh?
+
```
jimflores5: The words "Try this" beg for a repl link.  Playing with ``if`` statements is engaging and satisfying, especially since this is first chance the students have to really deviate from a proscribed path. ("What if I do THIS?  Huh.  That's cool.")

src/chapters/booleans-and-conditionals/nested-conditionals.rst
```diff
@@ -0,0 +1,68 @@
...
+Notice that when we put a conditional inside another conditional, the body of the nested conditional is indented by two tabs, and not just one. This convention allows us to easily determine which code is part of which conditional, by looking at the indentation level.
```
jimflores5: Notice that when we put one conditional inside another, the body of the nested conditional is indented by two tabs rather than one. This convention provides an easy, visual way to determine which code is part of which conditional.

src/chapters/booleans-and-conditionals/nested-conditionals.rst
```diff
@@ -0,0 +1,68 @@
...
+
+Check Your Understanding
+------------------------
+
+.. admonition:: Question
+
+   What is printed when the following code runs?
```
jimflores5: I like this example, but I feel that either one more concept check or a deeper problem would be beneficial.

1) Follow this concept check with a second, similar task.
OR
2) Start the code with "let num = x;" and then ask the students to identify the output for a set of possible x values.  Ex: 0, 2, 7, -5.
OR
3) As written, the code will never print "odd".  Instead of asking, "What is the output," retain only the conditional and ask, "Under what circumstances will the program print 'odd'?"

docs/_sources/chapters/booleans-and-conditionals/boolean-values-and-expressions.rst.txt
```diff
@@ -0,0 +1,138 @@
+Booleans
+========
+
+.. index:: data type
+
+One of the core features of any programming language is the ability to conditionally execute a segment of code. By this we mean that a program may run a segment of code *only if* a given condition is met, or if the program is in a particular state. 
+
+.. todo:: consider examples using bank app or messaging app
```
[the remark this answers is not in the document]
chrisbay: Fixed.

docs/_sources/chapters/booleans-and-conditionals/boolean-values-and-expressions.rst.txt
```diff
@@ -0,0 +1,138 @@
...
+
+.. admonition:: Example
+   
+   Consider a to-do list application that stores tasks along with their due dates and statuses (completed or uncompleted). The application may wish to notify the user that a task is due soon, but *only if* it has not already been completed.
+
+In this example, we can phrase that condition as follows: Send a notification of an upcoming due date if the statement "the task is uncompleted" is true.
```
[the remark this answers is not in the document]
chrisbay: This is more clear. I've updated it.

docs/_sources/chapters/booleans-and-conditionals/boolean-values-and-expressions.rst.txt
```diff
@@ -0,0 +1,138 @@
...
+
+.. index:: ! true, ! false, ! boolean
+
+.. index::
+   pair: boolean; value
+
+Boolean Values
+--------------
+
+There are only two **boolean values**. They are ``true`` and ``false``. As always, JavaScript is case-sensitive, so ``True`` and ``False`` are not valid boolean values.
```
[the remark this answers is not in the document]
chrisbay: Yep!

src/chapters/booleans-and-conditionals/booleans.rst
chrisbay: Good suggestion.

src/chapters/booleans-and-conditionals/booleans.rst
chrisbay: The difference between "related to" and "depending on" is subtle, but intentional. In JS, `==` doesn't really align with the general concept of equality, so I chose to say "related to." We introduce the difference between `==` and `===` later on, but at this point I don't want to make a statement that could be misleading, even if students are unlikely to pick up on it. Thoughts on other ways to rephrase to make the statement less awkward?

src/chapters/booleans-and-conditionals/booleans.rst
chrisbay: Good suggestion. I made "Comparison Operators" a level-3 heading.

src/chapters/booleans-and-conditionals/booleans.rst
chrisbay: I added this styling issue to [Trello](https://trello.com/c/8pkRJaXt). The title is a mandatory field for the RST directive `list-table`. We could hide such titles universally, but it strikes me that in other cases they may be necessary and useful. I'm inclined to leave them as-is for now, but we can raise the issue at a broader scale later on, if need-be.

src/chapters/booleans-and-conditionals/booleans.rst
chrisbay: done

src/chapters/booleans-and-conditionals/booleans.rst
jimflores5: Since ``==`` vs. ``===`` is discussed later, how about keeping the concrete "returns ``true`` or ``false`` depending on...", but using ``>`` or ``<`` instead of ``==``?

src/chapters/booleans-and-conditionals/booleans.rst
chrisbay: I placed this in a Question admonition.

src/chapters/booleans-and-conditionals/booleans.rst
chrisbay: I have come around to your original suggestion on this. There are a few other places on this page where we talk about `==` and it would be very awkward to say "depending on..." in each case. Additionally, there is a warning admonition that hints at the underlying reality.

src/chapters/booleans-and-conditionals/equality.rst
chrisbay: Fixed.

src/chapters/booleans-and-conditionals/equality.rst
chrisbay: Yeah, that's better. Fixed.

src/chapters/booleans-and-conditionals/equality.rst
chrisbay: ✅

src/chapters/booleans-and-conditionals/booleans.rst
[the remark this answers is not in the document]
chrisbay: Good point. I put this in a note admonition, but it also strikes me that there's an argument to standardize use of a generic admonition for cases like this (like we have done with "Try It!" and "Question" admonitions). Blocking it out in an admonition is a good way to communicate that it's different content, but it also doesn't fit our usual usage of Note, Tip, etc. What do you think?

src/chapters/booleans-and-conditionals/booleans.rst
```diff
@@ -0,0 +1,209 @@
...
+--------------
+
+.. index:: ! true, ! false, ! boolean
+
+.. index::
+   single: boolean; value
+
+The JavaScript data type for storing true and false values is ``boolean``, named after the British mathematician, George Boole. George Boole created `Boolean Algebra <https://en.wikipedia.org/wiki/Boolean_algebra>`_ which is the basis of all modern computer arithmetic.
+
+There are only two **boolean values**. They are ``true`` and ``false``. As always, JavaScript is case-sensitive, so ``True`` and ``False`` are not valid boolean values.
```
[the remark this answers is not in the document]
chrisbay: I structured this a bit differently to avoid starting a sentence with ``true``, which feels odd. It should be more clear now though.

src/chapters/booleans-and-conditionals/booleans.rst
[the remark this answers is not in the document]
chrisbay: This has been noted on the sphinx style issues card, to be addressed later.

src/chapters/booleans-and-conditionals/logical-operators.rst
```diff
@@ -0,0 +1,273 @@
...
+
+.. index::
+   single: operator; boolean
+
+JavaScript's logical OR operator, ``||``, also creates compound boolean expressions. This operator takes two operands, and the resulting expression is ``true`` if *either* of the operands are ``true`` individually. If both operands are ``false``, the overall expression is ``false``.
+
+As with logical AND, logical OR mirrors our experience of English language truth values. The sentence "Pigs can fly or dogs can run," is true as a whole. Joining the two clauses by "or" requires that only one of them is true in order for the full sentence to be true. Similarly, "Pigs can fly or dogs can speak Spanish," is a false statement.
+
+Let's look at some examples in JavaScript.
+
```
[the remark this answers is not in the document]
chrisbay: Agreed. I also made the similar example in the AND section an example.

src/chapters/booleans-and-conditionals/conditionals.rst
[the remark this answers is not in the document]
chrisbay: I used portions of both of these suggestions.

src/chapters/booleans-and-conditionals/logical-operators.rst
chrisbay: Since we aren't covering the bitwise operators, `&` and `|`, I'm not sure what an example would show. Perhaps the point could be better made with by emphasizing or rephrasing portions? The main points here are:
1. `&` and `&&` are different, as are `||` and `|`
2. Don't use `&` or `|`

src/chapters/booleans-and-conditionals/logical-operators.rst
chrisbay: Per previous discussion about table titles, I'm going to leave it but rephrase both the title and instructions to reduce redundancy.

src/chapters/booleans-and-conditionals/conditionals.rst
chrisbay: Your rephrasing could be read as implying that this is a special boolean expression that evaluates to `true` or `false`, while others may not. I think a parenthetical statement will clean this up while remaining precise.

src/chapters/booleans-and-conditionals/conditionals.rst
```diff
@@ -0,0 +1,233 @@
...
+
+      x is less than y
+
+Let's summarize the flow of execution of this conditional:
+
+#. Line 4 begins the conditional. The boolean expression ``x > y`` evaluates to ``false``, since 10 is not greater than 20. This causes line 5 to be skipped.
+#. Line 6 contains an else-if statement. The boolean expression ``x < y`` evaluates to ``true``, since 10 is less than 20. This triggers the execution of line 7.
+#. The code block associated with the ``else`` clause on lines 8-10 is skipped, because one of the conditions above was true.
+
+As with a simple ``if`` statment, the ``else`` clause is optional in this context as well. The following example does not print anything, since both conditions evaluate to false and there is no ``else`` clause.
```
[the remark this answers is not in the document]
chrisbay: Typo fixed. The statement is explicitly referring to the `else` clause (not `else if`) so the original version should remain.

src/chapters/booleans-and-conditionals/conditionals.rst
```diff
@@ -0,0 +1,233 @@
...
+   }
+
+We can construct conditionals using ``if``, ``else if``, and ``else`` with a lot of flexibility. The only rules are:
+
+#. We may not use ``else`` or ``else if`` without a preceding ``if`` statement.
+#. ``else`` and ``else if`` clauses are optional.
+#. A conditional may contain mutliple ``else if`` statements, each of which must follow the ``if`` statement and precede the ``else`` clause, if one is present.
+#. Only one ``else`` clause may be used.
+
+Regardless of the complexity of a conditional, *only one* of the code blocks that it contains will be executed.
```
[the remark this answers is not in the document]
chrisbay: Rephrased to "no more than one..."

src/chapters/booleans-and-conditionals/nested-conditionals.rst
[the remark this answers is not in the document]
chrisbay: I resolved this by removing the ! characters completely.

src/chapters/booleans-and-conditionals/logical-operators.rst
```diff
@@ -0,0 +1,273 @@
+Logical Operators
+=================
+
+.. index:: operators
+
+Recall that an operator is one or more characters that carries out an action on its operand(s). In :ref:`data-and-variables` we learned about three types of operators:
```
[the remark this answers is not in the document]
chrisbay: I am of the school of "use italics for emphasis sparingly." Convince me otherwise ;)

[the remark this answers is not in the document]
jimflores5: *Why* would you *want* to use *italics* *SPARINGLY*??!?!?!?

src/chapters/booleans-and-conditionals/booleans.rst
```diff
@@ -0,0 +1,209 @@
...
+
+      false
+
+.. index::
+   single: operator; comparison
+
+The ``==`` operator is one of six common **comparison operators**.
+
+.. index:: ==, ! !=, ! <, ! >, ! <=, ! >=
+
```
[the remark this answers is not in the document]
chrisbay: I added another column for such examples. They end up being mostly the "reverse" of the true examples, so let me know if this feels redundant.

src/chapters/booleans-and-conditionals/booleans.rst
chrisbay: There sort of is, in the form of a Try It! block with a Question on the same page. I do wonder if we are using these two formats together effectively, and how we should be differentiating them. I put this on the agenda for our next meeting.

src/chapters/booleans-and-conditionals/nested-conditionals.rst
chrisbay: Good idea. I added  a second question here that expands on the first.

src/chapters/booleans-and-conditionals/nested-conditionals.rst
chrisbay: Done!

src/chapters/booleans-and-conditionals/booleans.rst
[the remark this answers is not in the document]
gildedgardenia: I like that! That has been what I have been doing with extra cool, but not necessarily necessary information.

src/chapters/booleans-and-conditionals/logical-operators.rst
jimflores5: Ah, now I see:
let num = 55, num2 = 48;

console.log(num > 50 && num2 < 100);  Prints ``true`` or ``false`` as num & num2 change.
console.log(num > 50 & num2 < 100);  Prints 1 or 0 as num & num2 change.

The difference is subtle.  The question is whether ``true`` vs. 1 change the behavior of an if/else statement, a while loop or a return value.  My guess is that in most cases (especially novice coding), the results will be the same?

Perhaps the end statement could be made more specific.  How about, "The resulting value of the expression will be ``1`` or ``0`` instead of ``true`` or ``false``."  Then follow with some reasoning for why T/F is the way to go.

Alternative: "The single symbols ``&`` and ``|`` are themselves valid JavaScript operators, but using them produces values of ``1`` or ``0`` instead of ``true`` or ``false``.  If your code expects a boolean, then use ``&&`` and ``||``.  Using ``&`` or ``|`` MIGHT work, but it opens your code up for errors."

src/chapters/booleans-and-conditionals/conditionals.rst
jimflores5: Ok.

src/chapters/booleans-and-conditionals/booleans.rst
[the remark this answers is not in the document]
chrisbay: Nice. I created a custom "Fun Fact" admonition. I'll add an agenda item for our next meeting to discuss standardizing custom admonition usage.

PR #4, opened 2019-03-26
chapter 5 exercises

welzie:
- The content needs to be reviewed
- The style/layout also needs help (It seems very unreadable to blake)
- There is also one left to implement about debugging
- We should also possibly add more about `==` and `===`


src/chapters/booleans-and-conditionals/exercises.rst
```diff
@@ -0,0 +1,140 @@
+Chapter 5 Exercises
+===================
+
+Attempt these excersises to test your understanding.
```
gildedgardenia: exercises

src/chapters/booleans-and-conditionals/exercises.rst
```diff
@@ -0,0 +1,140 @@
+Chapter 5 Exercises
+===================
+
+Attempt these excersises to test your understanding. 
+Don't worry if you struggle while working on them. Struggling and recalling
+the material will make you remember it.
```
gildedgardenia: will help you remember it

src/chapters/booleans-and-conditionals/exercises.rst
```diff
@@ -0,0 +1,140 @@
...
+
+| Be sure to ask about the topics you don't understand in class, you won't be the only person that needs help with that topic.
+    
+.. note::
+
+   When the term **print** is used, that means to output a value to the ``console`` 
+   using ``console.log("text or a variable name")``.  
+   
+   | Print ``"hello"`` means ``console.log("hello")``
```
gildedgardenia: Delete this line, too redundant.

src/chapters/booleans-and-conditionals/exercises.rst
```diff
@@ -0,0 +1,140 @@
...
+   * - shuttleCabinReady
+     - true
+   * - crewStatus
+     - spaceSuitsOn && shuttleCabinReady
+   * - computerStatusCode
+     - 200
+   * - shuttleSpeed
+     - 15000
+
+Read the below code. What will be printed to the console?
```
gildedgardenia: I think this should be a list. The headings look odd.

src/chapters/booleans-and-conditionals/exercises.rst
```diff
@@ -0,0 +1,140 @@
...
+   } else if (engineIndicatorLight === "green blinking") {
+      console.log("engines are preparing to start");
+   } else {
+      console.log("engines are off");
+   }
+
+Write conditional expressions to satisfy the safety rules below
+----------------------------------------------------------------
+Using the variables defined from the table above.
+
```
gildedgardenia: I would remove one of the parts and make this a two part problem so that these exercises don't get too long.

src/chapters/booleans-and-conditionals/exercises.rst
```diff
@@ -0,0 +1,140 @@
...
+.. code-block:: javascript
+
+    if (!crewStatus || computerStatusCode !== 200 || !spaceSuitsOn) {
+        console.log("WARNING. Not ready");        
+    } else {
+        console.log("all systems go");
+    }
+
+Implement logic for spectrometer
```
gildedgardenia: This problem is too much. I would make this about the table above as opposed to introducing a spectrometer and a whole new set of conditions.

src/chapters/booleans-and-conditionals/exercises.rst
```diff
@@ -0,0 +1,140 @@
...
+
+  * print ``"Sodium found"``
+
+* else if ``colorCode`` is ``"Z491"`` OR ``waveLength`` is ``656`` OR ``486`` OR ``434`` nanometers.
+
+  * print ``"Hydrogen found"``
+
+* else print ``"Non target element found"``
+
+Fix the fuel status system
```
gildedgardenia: I would remove the debugging element from this problem, have people code this series of statements, and remove the above problem.

jimflores5: I like the idea of having the students evaluate someone else's code.

Maybe we could have the students compare two code blocks.  The first will begin with the comparison to 20000 and proceed down through 5000 (then the ``else`` clause), while the second block begins with the comparison to 5000 and works its way up to 20000 (then ``else``).

This will allow the students to consider how the order of the if statements affects the behavior of the code.  ("IF we want this behavior, HOW should we structure the code?")

If we want the students to code, we could follow the evaluation above with a task asking them to structure if statements that DO begin with 5000 yet still produce the desired output (e.g. the students could replace the ``>`` signs with ``<`` signs).

src/chapters/booleans-and-conditionals/exercises.rst
welzie: I feel very strongly that more questions is better than less.  I also think all the questions do not have to relate to the first question or table. These last two questions are a bit harder and are aimed at higher functioning students that need more of a challenge.

src/chapters/booleans-and-conditionals/exercises.rst
```diff
@@ -0,0 +1,140 @@
...
+* variables defined by spectrometer
+
+  * ``colorCode``
+  * ``waveLength``
+
+* if ``colorCode`` is ``"2A109"`` OR ``waveLength`` is ``2985`` nanometers.
+
+  * print ``"Water found"``
+
+* else if ``colorCode`` is ``"482C0"`` AND ``waveLength`` is ``589.0`` OR ``589.6`` nanometers.
```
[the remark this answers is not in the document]
welzie: How can `wavelength` be equal to 2 values at the same time?

PR #5, opened 2019-03-27
Chapter 4 studio

welzie:


welzie:
@jimflores5 if this looks please provide approval.

jimflores5:
The studio seems OK as a staring point, but I think we need to add more tasks or increase the rigor for generating LC04.  I feel the current state will be too straightforward and too short for a studio session.

Perhaps we could add more items to LC04 that depend on the result of conditional statements?  Right now, it looks like 11 of the 12 items on the list are simple outputs of the declared variables.  Only the "Clear for takeoff" item requires a deeper analysis.

welzie:
@jimflores5 for some reason I can't respond to your comment left about the difficulty of this studio. I am very open to making it harder, but I would like to get feedback from other test users first.

Also please create a review and click approve if you think this is ready to be merged.

PR #1, opened 2019-03-15
Data and variables chapter

chrisbay:
This chapter is ready to be reviewed. 

The primary concerns are:

- Mistakes: grammar, spelling, formatting, etc.
- Conventions (documented in the [curriculum docs](https://education.launchcode.org/curriculum-docs/guide/style.html)
- Coverage of [learning objectives](https://github.com/LaunchCodeEducation/lc101-learning-objectives/blob/master/unit1/lesson01.md)

Sally is the primary reviewer on this, but I'm adding the whole team since I know everyone is interested in seeing the initial content. 

Any structural stuff should be addressed directly, while content-related issues (suggestions for better examples, learning objective not covered) should be flagged on this request for clarification and resolution by the team.

PR #9, opened 2019-04-05
Drafts loops chapter

chrisbay:
Ready for QA. Things to be particularly mindful of:

- Are the loop flow diagrams too big? 
- Does the content motivate the concepts well enough? Loops are universal, yet the examples we present seem a bit dry.
- Are the explanations of `for` vs. `while` loops clear? These were hard to write, since JavaScript's "traditional" `for` doesn't provide a clear distinction between `for` and `while` loops, in the sense of "only this type of loop can carry out this type of task."


jimflores5:
"Are the loop flow diagrams too big?"

Two things need to be balanced for image sizes:
1) Is there text before or after the diagram that should be read with the picture?  If so, and if the text gets pushed out of view in order to see the entire image, then the diagram is too large.
2) Is there text inside the image?  If so, then the diagram needs to be large enough to keep the text legible.  (Ex: When Blake and I hit command++ to zoom, the text in the image should be at or above the font size of the text outside the image).

jimflores5:
"Does the content motivate the concepts well enough? Loops are universal, yet the examples we present seem a bit dry."

The content presents loops well enough (taking in future grammar/typo/content suggestions), but I would suggest more repl.it links to encourage experimentation.  Present the students with prompts such as, "Try to change this loop to print the numbers in reverse order," or, "What happens if you change i=0 to something else?"

src/chapters/loops/iteration.rst
```diff
@@ -0,0 +1,43 @@
...
+   console.log(0);
+   console.log(1);
+   console.log(2);
+   console.log(3);
+   console.log(4);
+   // and so on...
+
+Not only is this highly repetetive, but it is also error-prone. Even if utilizing copy-paste functionality, the sheer volume of code makes it somewhat likely that we will make a simple mistake, such as skipping an integer or misspelling ``console``.
+
+This code is also hard to modify. If we want to make a conceptually simple change---such as printing all the way to 100, or only printing even numbers---then we are forced to update an immense amount of code. Programming languages provide tools that allow us to repeat a sequence of statements in a much simpler way that this.
```
jimflores5: Grammar: Remove "that this" from the end of the last sentence.

src/chapters/loops/iteration.rst
```diff
@@ -0,0 +1,43 @@
...
+Repeated execution of a sequence of statements is called **iteration**. This chapter explores two mechanisms that JavaScript provides to make iteration simple and flexible---the ``for`` and ``while`` loops.
+
+To give you a taste of what's to come, here is how we could write the program above using a ``for`` loop.
+
+.. sourcecode:: js
+
+   for (let i = 0; i < 51; i++) {
+      console.log(i);
+   }
+
```
jimflores5: Many times, I've seen examples where the ‘let’ in the 1st statement in the () is omitted, and JS infers its usage.  How strict do we want to be here?  Dropping the ‘let’ cleans up the code appearance for novices, but does it conflict with our style guidelines?

src/chapters/loops/for-loops.rst
```diff
@@ -0,0 +1,90 @@
+**for** Loops
+=============
+
+.. index::
+   single: loop; for
+   single: iteration; definite
+
+.. index:: ! for loop
+
+The ``for`` loop is the first JavaScript tool for iteration that we will explore. A **for loop** allows us to carry out **definite iteration**. Definite iteration is the process of repeating a specific task with a specific data set. In other words, when a ``for`` loop begins it can usually be determined exactly how many times it will execute: once for each item in the data set.
```
jimflores5: Clunky grammar.  Streamline to remove the need for “In other words”.

src/chapters/loops/for-loops.rst
```diff
@@ -0,0 +1,90 @@
...
+
+We have already seen the basic syntax of a ``for`` loop.
+
+.. sourcecode:: js
+
+   for (let i = 0; i < 51; i++) {
+      console.log(i);
+   }
+
+This program prints the integers 0 through 50. In the language of definite iteration, we can say that the operation that this loop carries out printing a value to the console, and its data set is the collection of integers 0...50.
```
jimflores5: Grammar.  How about, “This program prints the integers 0 through 50, one number per line. In the language of definite iteration, we say that the loop has a data set of 0-50, and its action is to print a value to the console.”

src/chapters/loops/for-loops.rst
```diff
@@ -0,0 +1,90 @@
...
+::
+
+   for (initial expression; loop condition; update expression) {
+      loop body
+   }
+
+Notice that in the first line, within parentheses, the components **initial expression**, **loop condition**, and **update expression** are separated by semicolons. Let's look at these components in detail.
+
+- The statement ``let i = 0`` is executed exactly once, at the *beginning* of loop execution. The variable ``i`` is the **loop variable**. 
+- The boolean expression ``i < 51`` is the **loop condition**. This condition is evaluated before each loop iteration, or repetition. If the condition is ``true`` then the loop executes again. If the condition is ``false`` then the loop ceases execution and the program continues to execute the code below the loop. In the example above, an equivalent condition would be ``i <= 50``. It is common to use ``<`` when writing loop conditions, so that the last value used in the loop is *one less than* then right-hand side of the ``<`` symbol.
```
jimflores5: Grammar and typos. How about, “The boolean expression ``i < 51`` is the **loop condition**. This condition is evaluated before each loop iteration. If the condition is ``true`` then the loop executes. If the condition is ``false`` then the loop ceases, and the program continues with the code that follows the loop. In the example above, an equivalent condition would be ``i <= 50``. It is common to use ``<`` when writing loop conditions, so that the last value used in the loop is *one less than* the right-hand side of the ``<`` symbol.”

src/chapters/loops/for-loops.rst
```diff
@@ -0,0 +1,90 @@
...
+#. Steps 2 through 4 are repeated, using the new value of ``i``. This continues until the loop condition evaluates to ``false`` in step 2, ending the loop. In this example, this occurs when ``i < 51`` is ``false`` for the first time. Since our update expression adds 1 after each iteration, this occurs when ``i`` is 51 (so ``51 < 51`` is ``false``). At that point, the loop body will have executed exactly 51 times, with ``i`` having the values 0...50.
+
+In general, we can visualize the flow of execution of a ``for`` loop as a flowchart.
+
+.. figure:: figures/for-loop-flow.png
+   :height: 700px
+   
+   Flow of execution of a ``for`` loop
+
+.. warning:: If the body of a ``for`` loop has only a single line of code, then it may be placed on the line below the ``for`` statement *without* enclosing brackets, like this:
```
jimflores5: Eliminate the warning box!  If we want to encourage the students to ALWAYS use {} in for loops, then we should not give them an out.  We want to avoid contradicting our own style guide.  Perhaps we can add the 1-line-for-loop exception in an appendix, but let’s keep it out of the main content.

src/chapters/loops/iterating-over-collections.rst
```diff
@@ -0,0 +1,86 @@
+Iterating Over Collections
+==========================
+
+One of the most common uses of a ``for`` loop is to carry out a task once for each item in a collection. We have learned about two types of collections, strings and arrays. When using a loop with a collection in this way, we say that the loop *iterates over* the collection.
```
jimflores5: Bold or italics for “collections”, “strings” and “arrays”.

src/chapters/loops/iterating-over-collections.rst
```diff
@@ -0,0 +1,86 @@
+Iterating Over Collections
+==========================
+
```
jimflores5: Should we introduce the ``for (entry in string/array)`` concept in this section?

src/chapters/loops/for-statement-breakdown.rst
```diff
@@ -0,0 +1,171 @@
...
+
+   .. sourcecode:: js
+   
+      let name = "LaunchCode";
+
+      for (let i = 6; i < name.length; i++) {
+         console.log(name[i]);
+      }
+
+To avoid confusion and bugs, you should give your loop variable a unique name, one that you have not used elsewhere in your program. In cases for which the loop variable is serving as a "counter" for iterations of a loop, it is conventional to use ``i`` for the variable name. In the case of nested ``for`` loops (loops inside of loops), the variables ``j``, ``k``, etc. are often used.
```
jimflores5: Replace "for which" with "where".

src/chapters/loops/for-statement-breakdown.rst
```diff
@@ -0,0 +1,171 @@
...
+
+You will accidentally write an infinite loop at some point; doing so is a right of passage for new programmers. When this happens, don't panic. Stop your program and figure out why your loop condition never became false. 
+
+Update Expression
+^^^^^^^^^^^^^^^^^
+
+.. index::
+   single: for loop; update expression
+
+The final component in a for loop definition is the **updated expression**, which executes after *every* iteration of the loop. While this expression may be anything, it most often updates the value of the loop variable.
```
jimflores5: Replace “updated expression” with “update expression”.

src/chapters/loops/accumulator-pattern.rst
```diff
@@ -0,0 +1,143 @@
...
+.. sourcecode:: js
+
+   let str = "accumulator";
+   let reversed = "";
+
+Here, ``reversed`` is our accumulator variable. Our approach to reversing the string will be to loop over ``str``, adding each subsequent character to the *beginning* of ``reversed``, so that the first character becomes the last, and the last character becomes the first.
+
+.. admonition:: Example
+
+   .. sourcecode:: js
```
jimflores5: Alternative method: set up a for loop to proceed backwards through the string.  Start with i = str.length - 1 (to account for 0 indexing).

src/chapters/loops/while-loops.rst
```diff
@@ -0,0 +1,199 @@
...
+      num = input.question('Invalid input. Please enter a positive number:');
+      num = Number(num);
+   }
+
+.. index::
+   pair: input; validation
+
+This example illustrates the additional flexibility provided by ``while`` loops. While we use ``for`` loops to iterate over fixed collections (a string, an array, a collection of integers), the ``while`` loop can be used to iterate in more general circumstances. For the input validation example, at runtime it cannot be determined how many times the loop will repeat.
+
+.. warning:: It is *technically* possible to write a ``while`` loop as a ``for`` loop, since ``for`` loops use a condition, and the initial and update expressions can be left empty. Doing so, however, results in confusing code. This should not be done under any circumstances!
```
jimflores5: Remove the warning box!  It only serves to confuse the point.  

If you must include a warning, do NOT provide sample code and then tell the students, “Ignore this.”  Instead, focus on the general vs. specific nature of ``while`` and ``for`` loops.  All squares are rectangles, but not all rectangles are squares.  Similarly, a ``while`` loop can perform the same actions as any ``for`` loop, but a ``for`` loop cannot easily perform the same actions as any ``while`` loop.  Discourage trying ``while`` ---> ``for`` translations.

How about, “Given its greater flexibility, a ``while`` loop can be written in place of any ``for`` loop.  However, the reverse is NOT true.  Trying to replace a ``while`` loop with a ``for`` loop requires more work and a confusing code structure, and your code will probably require extensive debugging.  If you find yourself thinking, ‘How can I replace this ``while`` with a ``for``,’ STOP!  Just use the ``while`` loop.”

src/chapters/loops/while-loops.rst
```diff
@@ -0,0 +1,199 @@
...
+
+   let i = 0;
+
+   while (i < 51) {
+      console.log(i);
+   }
+
+This is an infinite loop. The variable ``i`` is initialized to 0 never updated, so the condition ``i < 51`` will always be true.
+
+Terminating a Loop With **break**
```
jimflores5: Suggestion: Shift “Terminating a Loop” to a different rst file.  Since the idea applies for both ``for`` and ``while`` loops, it should not appear in the ``while`` specific file.

src/chapters/loops/while-loops.rst
```diff
@@ -0,0 +1,199 @@
...
+            break;
+         }
+
+      }
+
+While not used often, ``break`` can be quite helpful in some situations. Consider a situation where we are searching for a particular element in an array. (While we have seen that JavaScript has array methods that can carry out array searches, many programming languages do not.) 
+
+We can use a ``while`` loop to say, essentially, *while we have not reached the end of the array, continue iterating*. We can then include a ``break`` within a conditional check to say, *when we have found the element we are searching for, exit the loop*.
+
+.. sourcecode:: js
```
jimflores5: Confusing shift to a ``while`` loop.  The same action could also be done with a ``for`` loop.  Show examples of each, or show different uses of ``break`` - one for each type of loop.

src/chapters/loops/choosing-loops.rst
```diff
@@ -0,0 +1,22 @@
+Choosing Which Loop to Use
+==========================
+
+.. index::
+   single: iteration; definite
+   single: iteration; indefinite
+
+The ``for`` loop is typically used to iterate through a fixed set of values that can be determined before the loop executes. This is why it is often said that a ``for`` loop exhibits **definite iteration**.
```
jimflores5: Word reduction: “This is why we say a ``for`` loop exhibits **definite iteration**.”

docs/chapters/loops/choosing-loops.html
```diff
@@ -0,0 +1,127 @@
...
+<p>While we saw that any <code class="docutils literal notranslate"><span class="pre">for</span></code> loop can be written as a <code class="docutils literal notranslate"><span class="pre">while</span></code> loop, by manually creating and updating a loop variable, it is preferable to use a <code class="docutils literal notranslate"><span class="pre">for</span></code> loop when iterating over a collection or iterating a fixed number of times. Manually updating the loop variable in a <code class="docutils literal notranslate"><span class="pre">while</span></code> loop is more work for you, the programmer, and can lead to infinite loops if not handled properly.</p>
+<div class="section" id="check-your-understanding">
+<h2>6.7.1. Check Your Understanding<a class="headerlink" href="#check-your-understanding" title="Permalink to this headline">¶</a></h2>
+<div class="admonition-question admonition">
+<p class="first admonition-title">Question</p>
+<p>You want to write a program that will run until January 1, 2050, redgardless of when it is started. Which loop should you use?</p>
```
welzie: I feel this could be too hard for them to understand that a while could be checking the current date/time. What do you think about changing it or adding more context?

docs/_sources/chapters/loops/for-loops.rst.txt
```diff
@@ -0,0 +1,90 @@
...
+::
+
+   for (initial expression; loop condition; update expression) {
+      loop body
+   }
+
+Notice that in the first line, within parentheses, the components **initial expression**, **loop condition**, and **update expression** are separated by semicolons. Let's look at these components in detail.
+
+- The statement ``let i = 0`` is executed exactly once, at the *beginning* of loop execution. The variable ``i`` is the **loop variable**. 
+- The boolean expression ``i < 51`` is the **loop condition**. This condition is evaluated before each loop iteration, or repetition. If the condition is ``true`` then the loop executes again. If the condition is ``false`` then the loop ceases execution and the program continues to execute the code below the loop. In the example above, an equivalent condition would be ``i <= 50``. It is common to use ``<`` when writing loop conditions, so that the last value used in the loop is *one less than* then right-hand side of the ``<`` symbol.
```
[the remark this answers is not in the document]
jimflores5: I agree with the sentiment here.  The last sentence could be omitted without loss of understanding.  If we want to steer the students to ``<`` rather than ``<=``, then remove the last 2 sentences.

src/chapters/loops/iteration.rst
chrisbay: Correct, if no keyword is used then `let` is inferred. We haven't discussed a convention/style preference. I lean toward being explicit (obviously) since I think it makes it more clear to students that a variable is being declared there, and nothing fishy is happening behind the scenes. But I think this would be good to discuss. I'll slot it into our next meeting.

src/chapters/loops/for-loops.rst
chrisbay: I streamlined this paragraph, but left (most of) the last sentence. As opposed to Python (and some others) not _all_ `for` loops in JS are definite.

src/chapters/loops/for-loops.rst
chrisbay: Valid point. The thing I'm guarding against here is the instructor or TA who tells a student, "You don't really need those brackets all the time." I fear that if we bury this comment in an appendix, then this scenario will happen more readily.

src/chapters/loops/iterating-over-collections.rst
chrisbay: For emphasis? Or because they are vocab terms? If you mean the latter, then these terms have already defined in earlier sections, hence don't need to be bold.

If you mean for emphasis, I don't agree. The point seem to warrant it, and having four italicized words in a single paragraph should rarely, if ever, happen.

src/chapters/loops/iterating-over-collections.rst
chrisbay: I am ambivalent on this question. I'll add it to the agenda for our next meeting.

src/chapters/loops/accumulator-pattern.rst
chrisbay: That does work, but isn't how the accumulator pattern is typically used. So in this case, I prefer the approach written.

src/chapters/loops/while-loops.rst
chrisbay: Good call!

docs/_sources/chapters/loops/while-loops.rst.txt
```diff
@@ -0,0 +1,199 @@
...
+There is another JavaScript construct that can also be used for iteration, the ``while`` loop. The ``while`` loop provides a much more general mechanism for iterating. Like a ``for`` loop, it uses a condition to determine whether the loop body will continue to execute. Unlike a ``for`` loop, however, it does not have initial and update expressions.
+
+**while** Loop Syntax
+---------------------
+
+The general syntax of a ``while`` loop looks like this:
+
+::
+
+   while (expression) {
```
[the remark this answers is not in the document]
chrisbay: Good point.

docs/chapters/loops/choosing-loops.html
chrisbay: I can see this, since we haven't introduced dates and times. I'll try to think of another conceptual example. Let me know if you have one!

src/chapters/loops/for-loops.rst
jimflores5: On the other hand, how often do you code single line for loops?  I'd let the TA who says, "You don't always need {}," explain the details.

docs/_sources/chapters/loops/for-loops.rst.txt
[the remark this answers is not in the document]
chrisbay: Good stuff.

src/chapters/loops/iterating-over-collections.rst
jimflores5: Emphasized because they are vocab words.  Since they have already been defined, then the point I was making has been addressed.

src/chapters/loops/iterating-over-collections.rst
jimflores5: OK.

src/chapters/loops/accumulator-pattern.rst
jimflores5: OK, or we could challenge the students to think of TWO ways to accomplish the task - one with ++ iteration and one with --.

src/chapters/loops/while-loops.rst
chrisbay: This feels like a topic to work out at a larger scale. This type of thing comes up a lot with JS. We want to steer students in a certain direction without hiding things from them (both because they'll see this stuff online anyway, and because a pro dev would be expected to know many of these things). What approach can we use to hit both of these objectives?

src/chapters/loops/for-loops.rst
jimflores5: OK.

src/chapters/loops/while-loops.rst
jimflores5: OK.

src/chapters/loops/while-loops.rst
jimflores5: Steer without hiding?  Tough call.  Perhaps something like, "Best practice strongly encourages us to avoid _____, but you will probably see such situations as you examine code samples online.  To explore ____ and see how awkward it is, click **here** for examples."  

Then we put the examples (carefully selected to emphasize the point we want to make) in an appendix.  Maybe title it "NOT best-practices".  If our goal is to teach the students how to spot and avoid sloppy code, then the examples should show the nastiness as well as a cleaner option.  Perhaps end the examples with a repl.it that asks the students to convert sloppy code into something more elegant that works better and is more readable.

I have other thoughts that would be easier to discuss rather than type.  Let's chat.

src/chapters/loops/exercises.rst
```diff
@@ -0,0 +1,63 @@
...
+
+.. sourcecode:: javascript
+
+    for (i=0; i < 100; i++){
+        console.log("Repetition is a good thing.")
+    }
+
+Loops simplify repetitive tasks!
+
+``for`` Practice
```
jimflores5: Styling fix: Change the ``for`` in the heading to **for**.

src/chapters/loops/exercises.rst
```diff
@@ -0,0 +1,63 @@
...
+
+#. Initialize two variables to hold the string ``'LaunchCode'`` and the array ``[1, 5, ‘LC101’, ‘blue’, 42]``, then construct ``for`` loops to accomplish the following tasks:
+    | a. Print each element of the array to a new line.
+    | b. Print each character of the string - in reverse order - to a new line.
+
+#. Construct a ``for`` loop that sorts the array ``[2, 3, 13, 18, -5, 38, -10, 11, 0, 104]`` into two new arrays:
+    | a. One array contains the even numbers, and the other holds the odds.
+    | b. Print the arrays to confirm the results.
+
+``while`` Practice
```
jimflores5: Styling fix: Change the ``while`` in the heading to **while**.

docs/chapters/loops/choosing-loops.html
chrisbay: @welzie Here is my updated question: "You are asked to program a robot to move tennis balls from one box (Box #1) to another (Box #2), one-by-one. The robot should continue moving balls until Box #1 is empty, and balls may be added while the robot is doing its work. 

   Which type of loop should you use to write the program?"

Thoughts?

jimflores5: You mean besides having the word "while" inside the problem asking for which loop to use?  ;-)

I like the question, especially if it is followed by another where the better option is ``for``.

chrisbay: Ha! That is a very explicit hint, isn't it? I'll fix that and add another one with `for` as the correct answer.

chrisbay:
@jimflores5 Ready for final review. I made these changes:
- Improved diagrams by increasing font size and decreasing overall diagram size
- Improved while vs. for concept check(s)
- Added  new "Try It!" examples w/ repl.it links (in "Breaking Down...," "The Accumulator Pattern," and "While Loops")

docs/chapters/loops/choosing-loops.html
jimflores5: Looks good.

PR #14, opened 2019-04-19
Drafts Functions chapter

chrisbay:
I feel good about this one. My only concerns are:

- Are there enough examples?
- There are a lot of complicated concepts in this chapter. Do they come across clearly, and are they well-motivated? 
- Do we need more concept checks? There aren't any that cover the process stuff, or "why we use functions." Mostly, this is because these are harder to articulate as questions and aren't as explicit in the learning objectives.


src/chapters/functions/intro.rst
```diff
@@ -0,0 +1,98 @@
...
+
+Here are some functions that you have become familiar with:
+
+- ``console.log``
+- The type conversion functions: ``Number``, ``String``, and ``Boolean``
+- String and array methods, such as ``indexOf``
+
+.. note:: When introducing string and array methods, we noted that a **method** is a function that "belongs to" an object. This distinction is important to keep in mind, and will be explored in depth in a later chapter. For now, note that a method is a special type of function.
+
+Each of the functions we have used works in the same way. Using the function's name, followed by parentheses, we can *call* the function, resulting in some action being carried out. Sometimes, as with ``console.log``, we can provide input data within the parenteses, which the function will use to carry out its action.
```
jimflores5: Typo: parenteses to "parentheses".

src/chapters/functions/intro.rst
```diff
@@ -0,0 +1,98 @@
...
+
+   ::
+
+      Hello, World!   
+   
+In addition to providing input, we may also be given a value in return when calling a function. This is what happened when we have used the type conversion functions, for example. 
+
+.. admonition:: Example
+
+   Type conversion functions *return* a value, that can be used by the calling code. Often, we store the return value of a function in a variable.
```
jimflores5: I'm not sure a type conversion function provides the best example for returning a value.  num = Number("46") returns 46, true, but something like arr = "Rutabaga".split('') returns a result that is markedly different from what went into the function.

src/chapters/functions/intro.rst
```diff
@@ -0,0 +1,98 @@
...
+   single: function; machine
+
+A commonly-used analogy for describing the concept of a function is that of a machine that takes input, carries out some action, and gives back some result. This is known as the **function machine** analogy.
+
+.. figure:: figures/function-machine.png
+   :alt: A "function machine," consisting of a box which takes inputs, and from which output emerges.
+
+   The function machine
+
+If we want to use a function, we must provide it some input (if needed). It then carries out an action using the input provided, and returns a result. What happens in the middle, when the function is doing its work, is "inside the machine" in a sense. If we know how a function should behave, we simply provide input, and receive output. The rest is up to the function machine itself.
```
jimflores5: Rephrasing to reduce passive voice.  How about "... It carries out an action on the input and returns a result.  The action occurs within the function, or 'inside the machine'.  If we know the purpose of a function, we simply provide it input and receive the output.  The rest is up to the machine itself."

If you want to reinforce the black box idea, add "without worrying about how one produces the other" to the end of the second to last sentence.

src/chapters/functions/intro.rst
```diff
@@ -0,0 +1,98 @@
...
+
+.. note:: You may notice that a function like ``console.log`` doesn't seem to return anything. We will soon learn that *every* function returns a value, regardless of whether or not that value is used, or is even useful.
+
+The programming concept of a function is very similar to the concept of a mathematical function. For example, in high school algebra you learned about functions such as polynomials. These functions used a mathematical variable (the input of the function) and then carried out some procedure using the input, producing a numerical result.
```
jimflores5: This paragraph might be too abstract for the math-phobic in the audience.  
Perhaps:
"The programming concept of a function is very similar to the concept of a mathematical function. For example, in high school algebra you learned about functions like y = 4x + 7. These functions used a mathematical input ("x") and  carried out a procedure to return a numerical result ("y")."

src/chapters/functions/intro.rst
```diff
@@ -0,0 +1,98 @@
...
+
+   ::
+
+      f(x) = x² + 3x - 2
+
+   We can *call* the function by giving it a specific *input*:
+
+   ::
+
+      f(3) = 3² + 3*3 - 2 = 9 + 9 - 2 = 16
```
jimflores5: Perhaps choose "4" for the input in order to distinguish it from the "3" in 3x.

src/chapters/functions/creating-functions.rst
```diff
@@ -0,0 +1,425 @@
...
+   **Output**
+
+   ::
+
+      Hello, Lamar!
+
+
+In this example, ``name`` is a **parameter**. It is part of the function definition, and *behaves like a variable* that exists only within the function.
+
+The value ``"Lamar"`` that is use when we invoke the function on line 5 is an **argument**. It is a *specific value* that is used during the function call.
```
jimflores5: Typo: 'used' rather than 'use'.

src/chapters/functions/creating-functions.rst
```diff
@@ -0,0 +1,425 @@
...
+   ::
+
+      Hello, Lamar!
+
+
+In this example, ``name`` is a **parameter**. It is part of the function definition, and *behaves like a variable* that exists only within the function.
+
+The value ``"Lamar"`` that is use when we invoke the function on line 5 is an **argument**. It is a *specific value* that is used during the function call. 
+
+You should think of the difference between an parameter and an argument as being the same as that between a variable and a value. A variable *refers to* a specific value, just like a parameter *refers to* a specific argument when a function is called. Like a value, a variable is a concrete piece of data.
```
jimflores5: Typos: 'a parameter' rather than 'an parameter'.  Also, "...an argument is a concrete piece of data."

How about: "The difference between a *parameter* and an *argument* is the same as the difference between a *variable* and a *value*. A variable *refers to* a specific value, just like a parameter *refers to* a specific argument when a function is called. Like a value, an argument is a concrete piece of data."

src/chapters/functions/creating-functions.rst
```diff
@@ -0,0 +1,425 @@
...
+
+   What does the following function return?
+
+   .. sourcecode:: js
+
+      function repeater(str) {
+          let repeated = str + str;
+          console.log(repeated);
+      }
+
```
jimflores5: I'm not sure that the intent of this CC question clearly matches the answers, especially with the follow-up CC in the next box.  I'd do away with this CC and keep the next.

src/chapters/functions/creating-functions.rst
```diff
@@ -0,0 +1,425 @@
...
+   #. What does the following code *output* when executed?
+
+   .. sourcecode:: js
+
+      function repeater(str) {
+          let repeated = str + str;
+          console.log(repeated);
+      }
+
```
jimflores5: Perhaps add the line ``repeater("Bob");`` after the function, and then modify the answer choices to provide concrete options (e.g. ``Bob``, ``BobBob``, ``undefined``, etc.

src/chapters/functions/creating-functions.rst
```diff
@@ -0,0 +1,425 @@
+Creating Functions
```
jimflores5: There is a LOT of content on this page.  Perhaps we should consider splitting it in two?  Maybe start the next page with ``return`` statements?

jimflores5:
Objective checklist:

    Explain what a function is - *Done*.

    Write a function with no parameters, a single parameter, and multiple
        parameters - (Still checking on this)

    Call a function, with and without parameters - (Still checking on this)

    Understand the behavior of parameters that are not provided by the
        caller - (Still checking on this)

    Use a return statement to complete execution of a function - *Done*
    Use a return statement to provide a value to the function caller - *Done*
    Explain what a return statement does - *Done*
    Understand variable scope as it pertains to functions - (Still checking on this)
    Explain what variable shadowing is - (Still checking on this)
    Write a function that calls another function - (Still checking on this)
    Use properly descriptive names when defining functions - *Done*

src/chapters/functions/why-create-functions.rst
```diff
@@ -0,0 +1,67 @@
...
+.. todo:: add reference to modules chapter
+
+As you will soon learn, encapsulating behavior within a function makes it easy for you to share that behavior across files and even projects. This becomes a critically important feature when working on large programs that consist of dozens, or even hundreds of files of code, as all professional developers do regularly. 
+
+Functions Save Millions of Lives Every Day
+------------------------------------------
+
+Okay, not really. But the point is, functions are incredibly powerful tools that you will come to appreciate and find indispensable. Ask a professional programmer if they could do their job without functions, and the answer will be an emphatic "NO!" 
+
+While they are abstract and can be difficult to learn, we promise you that the work will be worth it.
```
jimflores5: Spin this.  Instead of affirming that functions are abstract or hard to learn, how about "While functions may seem abstract and difficult to learn at first, repeated practice will quickly lead to mastery.  We promise that your work will be worth it."

src/chapters/functions/intro.rst
chrisbay: I added an example using `split`.

src/chapters/functions/creating-functions.rst
```diff
@@ -0,0 +1,425 @@
...
+
+      6
+
+Notice that ``sumToN`` does not print anything; the output comes from the final line of the program, which prints the value *returned by* the function call ``sumToN(3)``. Now that we have return statements in our coding toolbox, we will very rarely print anything within a function. If we want to see the value returned by a function then we must print it *after* calling the function. 
+
+.. admonition:: Question
+
+   The function ``sumToN`` uses a pattern that we have seen previously. What is it called? 
+
+Using **return** is Optional
```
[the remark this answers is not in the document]
chrisbay: I think we chatted about this last week. Instead of using code formatting in headings, we decided to use bold to improve display and accessibility in most situations. I added a task to Trello to change the bold font color when these headings are displayed on a white bg, for better readability.

src/chapters/functions/creating-functions.rst
chrisbay: These questions are intended to ensure students can differentiate between return values and output.

src/chapters/functions/process.rst
```diff
@@ -0,0 +1,72 @@
...
+A Good Function-Writing Process
+===============================
+
+The function is the most complex JavaScript construct that we have seen. A function has more components to its syntax than conditionals or loops. In addition, these components can be used in more complicated ways than those other constructs.
+
+To avoid frustration and bugs, it's important to approach writing functions in an intentional, structured way. This is especially true as the functions that you write become more and more complex. 
+
+In this section, we outline what we think is the best approach. To provide concrete examples, we will consider a fictional function that is able to make a sandwich.
+
+Design Your Function
```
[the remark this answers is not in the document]
chrisbay: Good call!

src/chapters/functions/parameters-and-variables.rst
```diff
@@ -0,0 +1,141 @@
...
+Function Scope
+--------------
+
+.. index:: 
+   single: function; scope
+   single: scope
+
+The **scope** of a variable is the extent to which a variable is visible within a program. In other words, a variable's scope consists of all locations within a program at which it can be used. Introducing functions gives us one of our first examples of limited variable scope---a situation in which a variable is not visible throughout an entire program.
+
+In particular, a variable defined within a function is not visible outside of that function.
```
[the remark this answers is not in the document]
chrisbay: Good suggestion. I also added "with `let`..." because a global variable could be visible outside the function.

src/chapters/functions/naming-functions.rst
```diff
@@ -0,0 +1,94 @@
+Naming Functions
+================
+
+As with variables, choosing good, descriptive names for the functions you write is important. It makes your code more readable, and therefore more maintainable and more bug-resistent.
+
+Use Camel Case
```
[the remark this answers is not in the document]
chrisbay: [Camel case](https://en.wikipedia.org/wiki/Camel_case) is not spelled with camel case, oddly enough!

src/chapters/functions/naming-functions.rst
```diff
@@ -0,0 +1,94 @@
...
+   - ``isReady``
+
+As we noted earlier, for boolean functions it is conventional that their names start with "is" or "has" whenever possible. 
+
+Creating a verb/noun pair for a function name doesn't always make sense, but when it does, you should use this format to create a good, descriptive name.
+
+Use Descriptive Names
+---------------------
+
+We've said it before that you should use descriptive names, but now we want to expand on the point a bit. You should *prefer long, descriptive names over short, abbreviated names*. If you can read a function name and understand what it does from that along, that is ideal.
```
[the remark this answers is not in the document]
chrisbay: I change the second sentence to: "If you can read a function name and understand what it does from the name alone, then the function has a good name." Along with the example that follows, does that make it more clear?

src/chapters/functions/composing-functions.rst
```diff
@@ -0,0 +1,126 @@
...
+      for (let i = 0; i < str.length; i++) {
+         reversed = str[i] + reversed;
+      }
+
+      return reversed;
+   }
+
+.. todo:: add reference to the correct section of the string chapter
+
+This is the same algorithm that we used to reverse a string in a previous chapter.
```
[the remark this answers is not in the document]
chrisbay: I don't recall, but I'll make a note to look.

src/chapters/functions/composing-functions.rst
```diff
@@ -0,0 +1,126 @@
...
+   }
+
+   function makeLunch( /*parameters*/ ) {
+      makeSandwich( /*parameters*/ );
+      pourDrink( /*parameters*/ );
+   }
+
+Why is this better? Smaller functions are easier to debug, for one thing. And by separating single responsibilities into individual functions, we also make our code easier to read and more reusable. In looking at the ``makeLunch`` function, it is very clear what is going on. First, it makes a sandwich, then it pours a drink. 
+
+Were the ``makeLunch`` function to simply contain all of the code necessary to carry out *both* tasks, there would be no clear separation between one task and the other, and the only way we might describe the various sections of the larger function would be to use comments. And, :ref:`as we've discussed <comments_lie>`, comments should be a secondary option for explaining your code.
```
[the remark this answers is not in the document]
chrisbay: The ref is in the correct place, but there is a CSS issue that puts the content behind the navigation bar. This has been noted in Trello, to be fixed later.

src/chapters/functions/naming-functions.rst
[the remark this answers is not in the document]
jimflores5: OK.

chrisbay:
> Write a function with no parameters, a single parameter, and multiple
>     parameters - *Not really done.*  Function were defined with and without parameters, but there was no explicit discussion about this idea.

In the section "Creating Functions" we state, "A function may be defined with several parameters, or no parameters at all." Does this cover it, or do you think a deeper explanation is warranted?

> Call a function, with and without parameters - *Half done*.  Function calls with and without arguments were done, but there was no explicit discussion about this idea.
> 
> Understand the behavior of parameters that are not provided by the
>     caller - *Not done*.

I added a new section to the page "Function Input and Output," called "Arguments Are Optional." This covers these last two objectives explicitly.

PR #18, opened 2019-04-25
Cleans up master

chrisbay:
- Standardizes exercise section names
- Fixes function references to not use parentheses
- Fixes build error caused by the Studios section of the TOC
- Fixes `console.log` in page heading to be formatted using bold


jimflores5:
Approved!

PR #20, opened 2019-04-30
1st assignment

jimflores5:


src/assignments/candidateQuiz.rst
```diff
@@ -0,0 +1,134 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+#. Define variables for:
+
+   a. candidate's name
+   b. a quiz question (pick any question from the table in step 2)
+   c. the correct answer
+   d. the candidate's response
+
+#. Ask for the candidate's name
```
chrisbay: I like the 3-step breakdown, with sub-steps. Let's also instruct learners to run their code after completing _every_ sub-step.

src/assignments/candidateQuiz.rst
```diff
@@ -0,0 +1,134 @@
...
+     - "3"
+
+Step 3: Use Iteration to Ask All Questions
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Add one or more loops to your code to ask all the questions in the quiz.
+Collect and check all the candidate's answers, then calculate the final
+percentage.
+
+   [TODO: Poll Ed Team for opinions on whether to provide the % calculation
```
chrisbay: Hard to say. I feel like people should know this, but it would suck for them to get hung up on something like this that isn't related to the assignment's objectives.

src/assignments/candidateQuiz.rst
```diff
@@ -0,0 +1,134 @@
...
+.. note:: Checking for the correct answer should be case insensitive (e.g. "Orbit" is the same as "orbit").
+
+.. list-table::
+   :header-rows: 1
+
+   * - Question
+     - Answer
+
+   * - True or false: 5000 meters = 5 kilometers.
+     - "True"
```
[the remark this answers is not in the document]
jimflores5: Actually, all of the answers will be checked as strings.

Blake and I considered having the students to type conversions for numbers and booleans, but we decided to keep the answer checking simple - especially since we have them use the ``toLowerCase`` method.

src/assignments/candidateQuiz.rst
```diff
@@ -0,0 +1,134 @@
...
+   whether their answer was correct.
+
+.. note:: Make sure your small app works properly before moving on to step 2.
+
+Step 2: Use Arrays
+^^^^^^^^^^^^^^^^^^
+
+Now that your small app is working, expand it to deal with multiple questions.
+
+#. Fill two arrays with the questions and answers listed in the table below.
```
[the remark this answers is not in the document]
jimflores5: Nothing is too paranoid when it comes to student's being creative in their interpretation of instructions.  I rephrased the question to be more specific.

src/assignments/candidateQuiz.rst
jimflores5: The sub-steps are just variable definitions, so there is nothing much to test.  I did add ``console.log`` suggestions to items 2 & 3.

src/assignments/candidateQuiz.rst
jimflores5: My thoughts exactly.  I'll put in the equation.

PR #21, opened 2019-04-30
Functions exercises.

jimflores5:


src/chapters/functions/exercises.rst
```diff
@@ -0,0 +1,156 @@
...
+   d. `Build your first function here <https://repl.it/@launchcode/FunctionsExercises01>`__.
+
+|
+
+1. (Continued) Now test your function:
+
+   e. Outside of the function, define the variable ``str`` and initialize it with
+      a string (e.g. ``'Functions rock!'``).
+   f. Define a second variable and initialize it with
+      ``someNameThatIChose = myFunctionName(str);``.
```
chrisbay: I recommend adding `let` to the declaration. Some students may do this on their own, but others will likely copy/paste or type it verbatim.

jimflores5: How about just using the prompt ``myFunctionName(str)``?

PR #22, opened 2019-05-02
Making Headlines (HTML) Studio

chrisbay:
Migrated as-is from the old curriculum.


src/chapters/html/studio.rst
```diff
@@ -0,0 +1,60 @@
...
+========================
+
+Getting Ready: Developer Tools
+------------------------------
+
+As you've learned, debugging code is an essential part of coding. When it comes to working with web pages, browser developer tools are indispensable.
+
+In this course, we will mostly use Firefox as our browser. If you haven't installed Firefox, you can get it from `Mozilla <https://www.mozilla.org/en-US/firefox/new/>`_. Other browsers have good developer tools as well (Chrome, in particular) but we're going to stick with Firefox. If you are more comfortable with another browser, you'll find that most of the developer tools functionality exists in other browsers' dev tools as well.
+
+Firefox's developer tools contain *a lot* of functionality, and we will only begin to touch on it here, learning more of its functionality later in the unit. In this class, we'll learn about the tools available for working with HTML elements.
```
gildedgardenia: I cannot remember. Did we agree that this intro should be in the studio itself?

src/chapters/html/studio.rst
```diff
@@ -0,0 +1,60 @@
...
+
+Walkthrough
```
gildedgardenia: None of the other studios have walkthroughs. Maybe move to instructor handbook?

src/chapters/html/studio.rst
```diff
@@ -0,0 +1,60 @@
...
+
+Studio
+------
+
+Pick a news site (`The New York Times <https://www.nytimes.com/>`_, for example), and use your browser's developer tools to modify one of the main articles to use a picture and text of your choosing.
+
+Have fun with this (see our example below), but be respectful of others and avoid overtly critical political or social commentary.
+
+.. figure:: figures/nelly-lc.png
+   :alt: A screenshot of the New York Times website, with a fake article discussing the rapper Nelly's visit to LaunchCode.
```
gildedgardenia: Maybe considering Nelly's latest controversies, we should use a different example.

src/chapters/html/studio.rst
```diff
@@ -0,0 +1,60 @@
...
+
+   A Sample Fake Article
+
+.. warning:: 
+   
+   When linking to an image, pay attention to the protocol---either ``http`` or ``https``---of the site you are modifying, and of the image you are including. You can find the protocol at the beginning of the image URL.
```
gildedgardenia: These dashes could be replaced with commas.

src/chapters/html/studio.rst
chrisbay: We discussed talking about dev tools in the HTML and CSS chapters, which means we'll have to talk about FF at that point. We haven't gone back to add that content yet though. I'll make a card for this, and move this text to the card in case we want to use it.

src/chapters/html/studio.rst
chrisbay: I rephrased the content here so that it's oriented toward the student. Since we'll be introducing these concepts in a chapter, there won't need to be a walkthrough.

src/chapters/html/studio.rst
chrisbay: Good call. I'll swap this example for a new one.

src/chapters/html/studio.rst
gildedgardenia: Sounds good!

src/chapters/html/studio.rst
gildedgardenia: Sounds good!

src/chapters/html/studio.rst
gildedgardenia: I like the new photo!

PR #23, opened 2019-05-08
More functions

chrisbay:
Includes content by both Jim and I. He will review mine (first four sections, on anonymous functions and functions as values), and I will review his (on recursion).


src/chapters/more-on-functions/why-use-anonymous-functions.rst
```diff
@@ -0,0 +1,38 @@
...
+      }, 5000);
+
+Defining an anonymous function at the same time it is passed as an argument prevents it from being used elsewhere in the program.
+
+Additionally, in programs that use lots of functions---such as web applications, as you will soon learn---defining functions anonymously, and directy within a function call, can reduce the number of names you need to come up with. 
+
+Anonymous Functions Are Ubiquitous in JavaScript
+------------------------------------------------
+
+JavaScript programmers use anonymous functions *a lot*. Many programmers use anonymous functions with the same gusto as that friend of yours that puts hot sauce on *everything*. For these programmers, just because an anonymous function isn't need to solve a problem doesn't mean that it *shouldn't* be used to solve the problem. Avoiding code that uses anonymous functions is impossible, since they are used so widely by so many JavaScript developers.
```
jimflores5: Grammar: "...that friend of yours who puts hot sauce on *everything*."

New paragraph with typo fix: "Just because an anonymous function isn't needed to..."

Suggestion: "...since they appear EVERYWHERE."  (Hyperbole appropriate?)

src/chapters/more-on-functions/recursion-walkthrough.rst
```diff
@@ -0,0 +1,207 @@
...
+above, the "identical step" is just adding two strings together - ``newString``
+and the next entry in the array.
+
+Third, build a function to accomplish the small steps: Let's call the function
+``combineEntries``, and we will set an array as the parameter.
+
+.. sourcecode:: js
+
+   function combineEntries(arrayName){
+      //some code here
```
chrisbay: Consider making this `// TODO` to indicate that we'll be filling this in as we go. In other places in the book, we simply omit a function body with a comment like the one you're using here, indicating that there really is code there, but we just aren't showing it. That isn't the case here.

src/chapters/more-on-functions/recursion-walkthrough.rst
```diff
@@ -0,0 +1,207 @@
...
+| ``for`` and ``while`` loops end when a particular condition evaluates to
+   ``false``. In the examples above, these conditions are
+| ``i < arr.length`` and ``arr.length > 0``, respectively.
+
+With recursion, we do not know how many times ``combineEntries`` must be
+called. To make sure the code stops at the proper time, we need to identify a
+condition that ends the process. This is called the **base case**, and it
+represents the simplest possible task for our function.
+
+``if`` the base case is ``true``, the recursion ends and the task is complete.
```
chrisbay: Code formatting is not needed for "if" on this line and the next. The word is being used colloquially here, and not in explicit reference to the `if` statement.

src/chapters/more-on-functions/recursion-walkthrough.rst
```diff
@@ -0,0 +1,207 @@
...
+
+.. sourcecode:: js
+
+   function combineEntries(arrayName){
+      if (baseCase is true){
+         //solve last step
+         //end recursion
+      } else {
+         //solve next small step
+         //call combineEntries again
```
chrisbay: These two lines of comments feel off. Saying "solve the next step" implies that some resolution to the problem will be found, and _then_ the function calls itself. Rather than finding some solution, we are reducing the problem down into a smaller version of the same problem, and then the function calls itself.

src/chapters/more-on-functions/recursion-walkthrough.rst
```diff
@@ -0,0 +1,207 @@
...
+      }
+   }
+
+``arrayName.length === 1`` sets up the condition for ending the recursion
+process. If it is ``true``, the single entry gets returned, and the function
+stops. Otherwise, ``combineEntries`` gets called again.
+
+Check Your Understanding
```
chrisbay: One thing that isn't explicitly called out in these sections is how crucial the base case is to correctly writing a recursive algorithm. You mention that the process "stops" when the base is reached, but a reader is never called to consider what happens if the base case isn't properly set up.

src/chapters/more-on-functions/recursion-walkthrough.rst
```diff
@@ -0,0 +1,207 @@
...
+   #. ``arr.includes('i')``
+   #. ``arr.indexOf('i')===-1``
+   #. ``arr.indexOf('i') !== -1``
+
+   Try it `here <https://repl.it/@launchcode/BaseCaseCC01>`__.
+
+.. admonition:: Question
+
+   The **factorial** of a number (n!) is the product of a positive, whole number and
+   all the integers below it.
```
chrisbay: This definition isn't quite right, since integers can be negative, so "all integers below it" would include an infinite set of negative ints. The classic definition of factorial is purely symbolic, which avoids potential awkwardness with the math terms: `n! = 1 * 2 * 3*...* n-1 * n`.

jimflores5:
Objectives:
- Write an anonymous function *Done*
- Explain how an anonymous function can be called, even though it doesn’t have a defined name *Done*, but how to call anon funcs when they are not assigned to variables is there, but not explicitly pointed out.
- Explain common uses of anonymous functions *Done*, but a quick review at the end of the content would put the uses in one place.

- Understand and describe how functions in JavaScript can be used as values *Done*

- Assign anonymous and named functions to variables *Done*
- Write a function that uses a parameter as a function *Done* (but should be phrased as "...uses a function as a parameter").
- Call a function that takes another function as an argument *Done*

- Describe the concept of a recursive problem, including the definitions of base case and recursive case (*To be answered by Chris*)

- Explain how base cases are important in preventing infinite loops (*To be answered by Chris*)
- Write a recursive function (*To be answered by Chris*)

src/chapters/more-on-functions/recursion-walkthrough2.rst
```diff
@@ -0,0 +1,133 @@
...
+
+Now we are ready to take the next step.
+
+A Visual Representation
+------------------------
+
+For arrays with more than one entry, we need to make ``combineEntries``
+consider the first element and then *check what is left in the array*. If the
+rest of the array contains more than one item, ``combineEntries`` calls iteslf
+again and repeats the process with a smaller set of entries.
```
chrisbay: These two sentences are a little ambiguous. One can read "check what is left" and "if the rest of the array contains more than one" to apply only to an array with two or more elements. It can also be read so as to apply only to an array with three or more elements ("rest of" could be what is left after taking the first element off). In fact, the recursive call also happens when there is only one element in the array.

src/chapters/more-on-functions/recursion-walkthrough2.rst
```diff
@@ -0,0 +1,133 @@
...
+
+To help visualize what's going on, let's start with the base case ``['L']``:
+
+.. figure:: figures/base-case-recursion.png
+   :alt: Visual representation for the base case.
+
+Nothing complicated here.  ``combineEntries`` sees only one item in the array,
+so it returns ``'L'``.
+
+Now consider an array with two elements ``['L', 'C']``:
```
chrisbay: Punctuation (insert comma): "...array with two elements, ``['L', 'C']``:"

src/chapters/more-on-functions/recursion-walkthrough2.rst
```diff
@@ -0,0 +1,133 @@
...
+
+So how do we code the ``else`` statement in ``combineEntries``? Recall what
+needs to happen each time the statement runs:
+
+a. Select the first element in the array,
+b. Call ``combineEntries`` again with a smaller array.
+
+Bracket notation takes care of part a: ``arrayName[0]``.
+
+| For part b, remember that the :ref:`slice method <slice-examples>` returns
```
chrisbay: It seems that these manual breaks are being use to solve issues that should be handled with CSS (not enough padding between paragraphs) or an RST directive (wanting a code sample to display on a new line). 

For a style issue, let's document it in Trello (with file name and screenshot) and leave the markup as-is (without the fix). 

For the second part (if displaying `arr = ...` on a separate line is indeed your intention), use `::` or ``.. sourcecode:: js``.

src/chapters/more-on-functions/recursion-walkthrough2.rst
```diff
@@ -0,0 +1,133 @@
...
+      } else {
+         return arrayName[0]+combineEntries(arrayName.slice(1));
+      }
+   }
+
+Each time the ``else`` statement runs, it extracts the first element in the
+array with ``arrayName[0]``, then it calls itself with the remaining array
+elements (``arrayName.slice(1)``). For ``combineElements(['L', 'C', '1', '0', '1']);``:
+
+| a. First call: Combine ``'L'`` with ``combineEntries(['C', '1', '0', '1'])``.
```
chrisbay: Let's chat about this formatting and how to achieve what you are going for. We should avoid hacking the markup language to achieve effects that isn't built for. 

This is a sequential list, which calls for `#.` as the item markers. Manually inserting line breaks means that the generated HTML won't render as a `<ol>` or even a `<ul>`, which can create problems with styling and document reading in various circumstances. 

I know RST doesn't have lettered lists, so my question would be what is the impetus for using lettered lists, and how important is the reasoning in this case?

src/chapters/more-on-functions/recursion-walkthrough2.rst
```diff
@@ -0,0 +1,133 @@
...
+| e. Fifth call: Base case returns ``'1'``.
+
+| To get the final result, proceed *up the chain*:
+| e. Return ``'1'`` to the fourth call,
+| d. Return ``'01'`` to the third call,
+| c. Return ``'101'`` to the second call,
+| b. Return ``'C101'`` to the first call,
+| a. Return ``'LC101`` as the final result.
+
+See recursion in action `here <https://repl.it/@launchcode/RecursionExample01>`__.
```
chrisbay: Make this a [semantic link](https://education.launchcode.org/curriculum-docs/guide/style.html#linking).

src/chapters/more-on-functions/recursion-walkthrough2.rst
```diff
@@ -0,0 +1,133 @@
...
+   result.
+
+As we make the array longer, ``combineEntries`` calls itself more times. Each
+call evaluates a smaller and smaller section of the array until reaching the
+base case. This sets up a series of return events - each one selecting a
+single entry from the array. Rather than building ``'LC101'`` from left to
+right, recursion constructs the string starting with the base case and
+adding new characters to the front:
+
+| ``'1'``
```
chrisbay: This might be better formatted with a table, including an additional column to describe the functional call or recursive level.

src/chapters/more-on-functions/recursion-walkthrough2.rst
```diff
@@ -0,0 +1,133 @@
...
+         return arrayName[0];
+      } else {
+         //solve next small step
+         //call combineEntries again
+      }
+   }
+
+Now we are ready to take the next step.
+
+A Visual Representation
```
chrisbay: This section turned out really well!

src/chapters/more-on-functions/recursion-wrap-up.rst
```diff
@@ -0,0 +1,53 @@
...
+
+a. Fewer lines of code required to accomplish a task,
+b. Makes code cleaner and more readable.
+
+Drawbacks of Recursion:
+
+a. More abstract than using loops,
+b. Code is "more readable" only if the reader understands recursion.
+
+Why Do I Need To Know Recursion?
```
chrisbay: Nice job on this. I think it will motivate students and contextualize recursion appropriately.

src/chapters/more-on-functions/recursion.rst
```diff
@@ -0,0 +1,88 @@
...
+   **Output**
+
+   .. sourcecode:: js
+
+      17
+
+Of course, there is no need to write a function to add 5 to a value, but the
+example demonstrates calling a function from within another function.
+
+What is 'Recursion'?
```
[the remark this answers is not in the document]
jimflores5: Fixed.

src/chapters/more-on-functions/recursion.rst
```diff
@@ -0,0 +1,88 @@
...
+---------------------
+
+.. index:: ! recursion
+
+**Recursion** is the process of solving a larger problem by breaking it into
```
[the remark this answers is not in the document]
jimflores5: I recognize that a function calling itself is one of the key ideas in recursion.  I felt, however, that a small lead-in from "identical pieces" to "calling itself" would help.  About 3 sentences after the "pieces" idea, I bring in the "calling itself" part.

We can reformulate the intervening text, but I like the idea of tying recursion to something the students are familiar with, namely loops.  Since a loop performs a single task multiple times, I think it provides a good mental model for what recursion accomplishes.

src/chapters/more-on-functions/recursion.rst
```diff
@@ -0,0 +1,88 @@
...
+Solving the big problem involves completing and combining the smaller parts.
+We accomplish this by either:
+
+a. Setting up a loop to call one function lots of times, OR
+b. Using the strength of recursion.
+
+Here is the clever idea behind recursion.  Instead of a loop, a function simply
+calls *itself* over and over again until the larger problem is solved.
+
+.. admonition:: **Recursion in a nutshell:**
```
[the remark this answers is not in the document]
jimflores5: I considered that as well, but I wasn't sure how the headings "What Is Recursion?" and "Recursion in a Nutshell" would play off of each other.  Why read the intro when you can skip down to the bullet list?  I would prefer to force the intro reading.

I've added the heading so we can compare the two layouts.

src/chapters/more-on-functions/recursion-walkthrough.rst
```diff
@@ -0,0 +1,207 @@
+Recursion Walkthrough: The Base Case
+=====================================
+
+To ease into the concept of recursion, let's start with a loop task.
+
+In the Arrays chapter, we examined  the ``join`` method, which combines the
```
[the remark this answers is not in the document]
jimflores5: Done.

src/chapters/more-on-functions/recursion-walkthrough.rst
```diff
@@ -0,0 +1,207 @@
...
+         console.log(arr);
+
+       **Output**
+
+       .. sourcecode:: js
+
+         'LC101'
+         [ ]
+
+Inside each loop, the code simply adds two strings together - whatever is
```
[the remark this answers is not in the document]
jimflores5: Done.  Also replaced '-' with '---'.

src/chapters/more-on-functions/recursion-walkthrough.rst
jimflores5: Done.

src/chapters/more-on-functions/recursion-walkthrough.rst
jimflores5: I know, but I thought I'd apply a little artistic license here, especially since we need an ``if`` statement in the code.  You know, foreshadowing, a quality literary technique.  ;-)

``if`` you feel strongly about this, I'd be OK changing it back.

src/chapters/more-on-functions/recursion-walkthrough.rst
jimflores5: Fixed.  ``else`` now just calls ``combineEntries`` again.

src/chapters/more-on-functions/recursion-walkthrough.rst
jimflores5: Got it.  I've added a new section to deal with this idea.  I'll fill in the text and examples soon.

src/chapters/more-on-functions/recursion-walkthrough.rst
jimflores5: Google failed!  I wanted to be accurate, so I queried "definition of factorial", and this is what  lord Google told me.

I fixed the definition.

src/chapters/more-on-functions/recursion-walkthrough2.rst
jimflores5: It was the second part.  Splitting the code over two lines looked bad in this case.

Fixed.

src/chapters/more-on-functions/recursion.rst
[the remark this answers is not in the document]
jimflores5: Heading it is!

src/chapters/more-on-functions/anonymous-functions.rst
```diff
@@ -0,0 +1,123 @@
...
+      // function body
+
+   }
+
+You might be asking yourself, *How can I call such a function if it doesn't have a name?!* Good question. Let's address that now.
+
+Anonymous Function Variables
+----------------------------
+
+Anonymous functions are often assigned to variables when they are created, which allows them to be called using the variable's name.
```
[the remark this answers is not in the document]
chrisbay: There are two primary alternatives to putting an anonymous function in a variable:

1. Passing it directly to another function.
2. Doing this seemingly crazy, but sometimes useful thing:

```js
(function(){
    // body
})();

This creates the function and then immediately calls it. 
```

src/chapters/more-on-functions/anonymous-functions.rst
```diff
@@ -0,0 +1,123 @@
...
+
+.. admonition:: Example
+
+   Let's create and use a simple anonymous function that returns the sum of two numbers.
+
+   .. sourcecode:: js
```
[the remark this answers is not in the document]
chrisbay: Good question. Let's discuss at our next meeting.

src/chapters/more-on-functions/recursion-walkthrough2.rst
jimflores5: I was going for an "up the chain" visual. The calls to ``combineElements`` move from a --> b --> c --> d --> e, and I wanted to reinforce how LC101 is built from e --> d --> c --> b --> a.

As for lettered lists, I've been using them as sub-bullets for numbered exercises, or as stand-alone lists when I think letters provide more clarity than numbers.  This is simply a style choice.

src/chapters/more-on-functions/recursion-walkthrough2.rst
jimflores5: Fixed.

src/chapters/more-on-functions/recursion-walkthrough2.rst
jimflores5: Whew!

src/chapters/more-on-functions/recursion-wrap-up.rst
jimflores5: Thanks.

src/chapters/more-on-functions/recursion-walkthrough2.rst
jimflores5: I'm not sure what you are envisioning here.  Let's chat.

src/chapters/more-on-functions/recursion.rst
```diff
@@ -0,0 +1,88 @@
...
+
+   a. Split a big problem into small, identical pieces.
```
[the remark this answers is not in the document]
jimflores5: Fixed this by implementing your suggestions (kinda) and moving the "nutshell" section to the wrap-up page.

src/chapters/more-on-functions/recursion-walkthrough2.rst
jimflores5: Fixed.

src/chapters/more-on-functions/recursion.rst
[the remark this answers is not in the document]
jimflores5: d
ed
xed
ixed
Fixed

src/chapters/more-on-functions/recursion.rst
```diff
@@ -0,0 +1,88 @@
...
+function rather than coding several different functions.
+
+Solving the big problem involves completing and combining the smaller parts.
+We accomplish this by either:
+
+a. Setting up a loop to call one function lots of times, OR
+b. Using the strength of recursion.
+
+Here is the clever idea behind recursion.  Instead of a loop, a function simply
+calls *itself* over and over again until the larger problem is solved.
```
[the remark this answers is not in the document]
jimflores5: Fixed with new What Is Recursion text.

src/chapters/more-on-functions/recursion.rst
```diff
@@ -0,0 +1,88 @@
...
+**Recursion** is the process of solving a larger problem by breaking it into
+small, easy steps that *can all be solved in exactly the same way*.
+
+Splitting a large task into small, identical pieces allows us to reuse a single
+function rather than coding several different functions.
+
+Solving the big problem involves completing and combining the smaller parts.
+We accomplish this by either:
+
+a. Setting up a loop to call one function lots of times, OR
```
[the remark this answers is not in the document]
jimflores5: Fixed with new intro text.

src/chapters/more-on-functions/functions-as-values.rst
```diff
@@ -0,0 +1,67 @@
+Functions as Values
+===================
+
+.. index:: function; as a value
+
+As we have noted, functions are a powerful programming feature. JavaScript in particular allows functions to be used in flexible ways that are not possible in many other languages. This chapter introduces a bit more of the power of functions.
+
+Functions Are Data
```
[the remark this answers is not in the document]
chrisbay: Ha!

src/chapters/more-on-functions/functions-as-values.rst
```diff
@@ -0,0 +1,67 @@
...
+
+In particular, functions have a data type, just like all other values. Recall that a **data type** is a group of values that share characteristics, such as strings and numbers. All strings have a length, while numbers don't. Numbers can be manipulated in ways that strings cannot be, via operations like division and subtraction. 
+
+.. admonition:: Example
+
+   The data type of the type conversion function ``Number`` is ``function``. In fact, all functions are of type ``function``.
+
+   .. sourcecode:: js
+   
+      console.log(typeof Number);
```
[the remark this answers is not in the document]
chrisbay: Good call. I added statements for a number and a string, but left off an array since it reports its type as `object`. Since we haven't discussed this, it could be confusing.

src/chapters/more-on-functions/functions-as-values.rst
```diff
@@ -0,0 +1,67 @@
...
+
+When a variable refers to a function, we can call the function using the variable name:
+
+.. sourcecode:: js
+
+   helloFunc();
+
+The variable ``helloFunc`` can be thought of as an *alias* for the function ``hello``. When we use the name ``helloFunc``, JavaScript sees that it refers to the function ``hello`` and will use that *specific* function. 
+
+We are used to this behavior when a variable holds a number or a string. When a variable holds a function, the variable "points at" or "refers to" the function. When we use the variable name, it is as if we are using the specific value that the variable refers to.
```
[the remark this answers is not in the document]
chrisbay: I think this paragraph is better now. 

You got me on the scare quotes! That phrase deserves italics instead.

src/chapters/more-on-functions/anonymous-functions.rst
```diff
@@ -0,0 +1,123 @@
...
+.. admonition:: Question
+
+   Consider the code sample below, which declares an anonymous function
+   beginning on line 1.
+
+   .. sourcecode:: js
+      :linenos:
+
+      let f1 = function(str) {
+         console.log(str + str);
```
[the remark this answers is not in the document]
chrisbay: Good catch. I converted the `console.log` to a `return` statement.

src/chapters/more-on-functions/functions-as-arguments.rst
```diff
@@ -0,0 +1,285 @@
+Functions as Arguments
+======================
+
+You now know that *functions are data*, and therefore can be passed around just like other values. In particular, this means a function can be *passed to another function* as an argument. This turns out to be extremely useful. 
+
+Passing Functions as Arguments
+------------------------------
+
+Lots of JavaScript functions allow you pass a function as an argument when calling them. A function passed in this way can be called later, potentially multiple times. It allows the *function being called* to use the *function argument* to carry out its action.
```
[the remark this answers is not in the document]
chrisbay: I think the first sentence is better now. I removed the first "function" since using "argument" and "pass" implicitly make it clear what we're talking about.

The second sentence is not specifically about callbacks, though callbacks are used in this way. It attempts to explain what happens to the function argument, and how it is used by the receiving function. The third sentence explains this better though, so I'll axe the second.

src/chapters/more-on-functions/functions-as-arguments.rst
```diff
@@ -0,0 +1,285 @@
...
+
+      console.log(doubled);
+
+   **Output**
+
+   ::
+
+      [ 6.28, 84, 9622 ]
+
+Using Parameters as Functions
```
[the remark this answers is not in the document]
chrisbay: Who's on first ;)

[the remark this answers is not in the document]
chrisbay: Rephrased to: Functions That Take Function Arguments

src/chapters/more-on-functions/functions-as-arguments.rst
```diff
@@ -0,0 +1,285 @@
...
+-----------------------------
+
+The previous section illustrates how a function can be passed to another function as an argument. This section takes the opposite perspective to *write* functions that can take other functions as arguments.
```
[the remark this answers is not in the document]
chrisbay: I think we clarified this yesterday. If not, let's discuss further.

src/chapters/more-on-functions/functions-as-arguments.rst
```diff
@@ -0,0 +1,285 @@
...
+
+      function logError(msg, logger) {
+         logger('ERROR: ' + msg);
+      }
+
+      logError('Something broke!', fileLogger);
+
+This example can be made even more powerful by enabling multiple loggers.
+
+.. admonition:: Example
```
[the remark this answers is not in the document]
chrisbay: Regarding this and the comment immediately preceding: What if the validator and logger examples were broken off into a separate page with expanded explanation? Would that bring it down to an approachable level?

src/chapters/more-on-functions/functions-as-arguments.rst
```diff
@@ -0,0 +1,285 @@
...
+            logger[i]('ERROR: ' + msg);
+         }
+
+      }   
+
+      logError('Something broke!', [fileLogger, consoleLogger]);
+
+As with the validation example, these programs separate behaviors in a way that makes the code more flexible. To add or remove a logging destination, we can simply change the way that we call ``logError``. The code *inside* ``logError`` doesn't know or care about the specfics of how each logging function does it's job. It is concerned merely with creating the message string and passing it to each logger.
+
+When writing a function that uses one of its parameters as a function, things can go wrong since there is now way to *force* a programmer to pass a function.
```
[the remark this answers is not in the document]
chrisbay: I converted this chunk to be its own section, titled "A Word of Caution." It felt weird to me to use "*Caution:*" over the warning admonition, while using the latter would have resulted in three admonitions in a row with not text in between (also not great). I think the heading sets off the content appropriately. 

What do you suggest as an alternative to "a programmer"? I avoided using "us" or "you" because I wanted to emphasize that we don't just write functions for ourselves, but for others to use too.

src/chapters/more-on-functions/why-use-anonymous-functions.rst
chrisbay: The hyperbole feels a bit strong. I shortened this to, "Avoiding JavaScript code that uses anonymous functions is impossible." Better?

src/chapters/more-on-functions/recursion-walkthrough2.rst
chrisbay: Regarding lettered lists: I'm not opposed to their use if they result in semantic HTML. I don't think Sphinx does this out of the box, but there may be a plugin that could enable the behavior, since CSS supports letters for `ol` elements.

It would also be good to discuss conventions, if we're able to get lettered lists to work properly. Mixing numbered and lettered lists for the same use cases (e.g. exercise sub tasks) would be sloppy.

src/chapters/more-on-functions/functions-as-arguments.rst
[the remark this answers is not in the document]
jimflores5: I don't know is on third, and let's not forget about Why and Because.  :-)

src/chapters/more-on-functions/functions-as-values.rst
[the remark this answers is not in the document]
jimflores5: The first paragraph is much better now, as is the second.  +1

src/chapters/more-on-functions/functions-as-arguments.rst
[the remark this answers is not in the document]
jimflores5: The separate section works well.  As for avoiding the programmer, maybe we could focus on the behavior of the code/function itself.  Something like, "Whenever a function uses one of its parameters as a function, errors occur if something other than a function gets passed."  I'm not happy with using 'function' 3 times in the sentence, but maybe we can adjust from here.

src/chapters/more-on-functions/why-use-anonymous-functions.rst
jimflores5: Yep.  Better.

src/chapters/more-on-functions/recursion-walkthrough2.rst
jimflores5: I mix numbers and letters for clarity in exercises.

Ex:
1. Big problem statement.
  a. First sub-task,
  b. Second sub task.

2. New problem.
  a. Etc.
  b. Etc.

I think this works better than:
1. Big problem statement.
  1. First sub-task,
  2. Second sub task.

2. New problem.
  1. Etc.
  2. Etc.

src/chapters/more-on-functions/recursion-walkthrough2.rst
jimflores5: Fixed the lettered list hack and put the 3 sets of items into list tables.  Let me know what you think.

src/chapters/more-on-functions/recursion-walkthrough2.rst
chrisbay: Makes sense. Interestingly, the RST docs on the Sphinx site don't cover lettered lists (to my knowledge). As long as they result in `ol` elements in the HTML, I'm happy. 

Maybe we can make your practice a convention then: Ordered sublists of ordered lists are lettered.

src/chapters/more-on-functions/functions-as-arguments.rst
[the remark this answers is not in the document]
chrisbay: I modified this page heavily, breaking it into two pages and adding additional explanation to the logging example. Let me know what you think.

src/chapters/more-on-functions/passing-functions.rst
```diff
@@ -0,0 +1,149 @@
...
+   
+      let nums = [3.14, 42, 4811];
+
+      let timesTwo = function (n) {
+         return n*2;
+      };
+
+      let doubled = nums.map(timesTwo);
+
+      console.log(doubled);
```
jimflores5: How about having ``console.log(nums);`` as well, just to show that ``map`` does not mutate the original array?

src/chapters/more-on-functions/passing-functions.rst
```diff
@@ -0,0 +1,149 @@
...
+
+   `Halve the array at repl.it <https://repl.it/@launchcode/Arraymap-check>`_
+
+.. admonition:: Question
+
+   Use the ``map`` method to map an array of strings. For each name in the array, map it to the first initial.
+
+   .. sourcecode:: js
+
+      let names = ["Chris", "Jim", "Sally", "Blake"];
```
jimflores5: You forgot "Paul"!!!

src/chapters/more-on-functions/receiving-functions.rst
```diff
@@ -0,0 +1,210 @@
...
+      }
+
+      function logError(msg, logger) {
+         let errorMsg = 'ERROR: ' + msg;
+         logger(errorMsg);
+      }
+
+      logError('Something broke!', fileLogger);
+
+Let's examine this example in more detail.
```
jimflores5: This is much better than before. Your description of how the code works is clear and easy to follow, and it keeps these examples from being too heavy of a lift for the students.

src/chapters/more-on-functions/receiving-functions.rst
```diff
@@ -0,0 +1,210 @@
...
+            logger[i](errorMsg);
+         }
+
+      }   
+
+      logError('Something broke!', [fileLogger, consoleLogger]);
+
+The main change to the program is that ``logError`` now accepts an *array* of functions. It loops through the array, calling each logger with the message string.
+
+As with the validation example, these programs separate behaviors in a way that makes the code more flexible. To add or remove a logging destination, we can simply change the way that we call ``logError``. The code *inside* ``logError`` doesn't know about how each logging function does it's job. It is concerned merely with creating the message string and passing it to the logger(s).
```
jimflores5: Grammar: Remove the 'about' from "...doesn't know about how each logging function..."

Personal preference: "only" instead of "merely".

src/chapters/more-on-functions/functions-as-arguments.rst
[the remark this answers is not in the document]
jimflores5: This is a MUCH better approach.  Just a couple of typo/grammar corrections, which I noted in new comments.

src/chapters/more-on-functions/functions-as-arguments.rst
```diff
@@ -0,0 +1,285 @@
...
+
+Another common example of a function using another function to customize its behavior is that of logging. A real-world application is typically capable of logging messages such as errors, warnings, and statuses. Many such applications allow for log messages to be sent to one or more destinations. For example, the application may log messages to both the console and to a file.
+
+We can write a logging function that relies on a function parameter to determine the logging destination.
+
+.. admonition:: Example
+
+   The ``logError`` function outputs a standardized error message to a location determined by the parameter ``logger``.
+
+   .. sourcecode:: js
```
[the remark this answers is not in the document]
jimflores5: Fixed well by Chris B!

PR #24, opened 2019-05-08
Introduces brief and full TOCs

chrisbay:
- Reduces the depth of the Chapters section of the main TOC to 1
- Creates a new TOC in `index-full.rst` which is exactly the same as the old TOC
- Adds crosslinks to the top of each TOC 

I had to tweak the styles a bit for the smaller TOC, which I did in the `layout.html` template. I'm not super fond of this solution since it depends on the the id of Chapters heading on the main TOC, which in turn depends on the actual text in that heading. If we like this format, I'll find a better solution for adjusting the styles.


jimflores5:
Meh.  I could go either way.  I still like the drop-down idea to see chapter items in the short TOC, or having that content appear in a right-hand aside when the heading is clicked.

The only suggestion I have for the current template would be to remove the "Contents in Brief" section and replace it with a "Chapters (view full contents)" heading (link).  The full TOC would then have the heading "Chapters (view abbreviated contents)".

PR #26, opened 2019-05-08
Drafts functions studio

chrisbay:
This studio is adapted from an [activity](https://github.com/LaunchCodeEducation/problem-solving-practice/tree/master/shapes) that I wrote a couple of years ago for CodeCamp and hasn't been used much since.

I chose to work with this because it shows students how to write and compose small functions to solve larger problems. It seems potentially long for a studio, so we may need to mark some of the later parts of it as optional.


jimflores5:
I remember doing this activity in LC101 as part of the Python module.  It was a self-reinforcing process, since each task is short on its own and provides immediate feedback.

We should sit down and look at your studio and my exercises side by side.  There is some overlap between the two, and we could choose which one to make the studio and which one the exercises.

src/chapters/functions/studio.rst
```diff
@@ -0,0 +1,169 @@
+Studio: Shapes
```
jimflores5: Rename: "Studio: Functions" to be consistent with the other chapters.

This is a nice stroll down memory lane.  :-)

Pyyyyyyythoooooooonnnnnnnn!

src/chapters/functions/studio.rst
```diff
@@ -0,0 +1,169 @@
...
+     #####     
+      ###      
+       #
+
+There is no starter code for this studio, so create a new Node.js project at `repl.it <https://repl.it/student>`_ to get started.
+
+Rectangles
+----------
+
+#. Write a function ``line(size)`` that returns a line with exactly ``size``
```
jimflores5: So Python has ``print('#'*size)`` to get a line of #s (``print('#'*5)`` = '#####').  Is there a console.log way to do this, or do the students need to use the ``repeat`` string method?  Is ``repeat`` on the string methods page?

src/chapters/functions/studio.rst
```diff
@@ -0,0 +1,169 @@
...
+   
+          #
+         ###
+        #####
+       #######
+      #########
+
+   .. tip:: Consider the top line of the triangle to be level 0, the next to be line 1, and so on. Then line ``i`` is a space-line with ``height - i - 1`` spaces and ``2 * i + 1`` hashes.
+
+Diamonds
```
jimflores5: Students who get this will have no trouble finishing the entire studio in the allotted time.  Students who struggle will hit the wall around the isosceles triangle.

I'd classify "Diamonds" as a bonus mission, and merge the "use characters other than '#'" missions into a single task.  This could go above the diamonds task in the bonus missions, or it could be slipped in after the ``line`` task.

src/chapters/functions/studio.rst
chrisbay: In my solution I use a `for` loop with the accumulator pattern, since that's what I imagined most students would do. I wouldn't expect them to know `String.repeat`, and no, there isn't a version of the `*` operator for this in JS like there is in Python.

src/chapters/functions/studio.rst
chrisbay: Makes sense. I thought that it felt a little big. I'm happy to make diamonds a bonus mission.

Is there a better way to point students in the right direction on the triangle problem? Or do they just need to struggle with it?

chrisbay:
@jimflores5 I implemented all of your suggest changes, but will wait until we have a chance to chat about studio vs. exercises before making final updates and merging.

src/chapters/functions/studio.rst
jimflores5: Struggle.  As the last task before the diamonds, it is well placed to push the students outside of their comfort zone, but not so far that the solution is out of reach.

PR #29, opened 2019-05-15
Opens external links in new tab

chrisbay:
I added some JS to add `target="_blank"` to any anchor tag not linking to either a launchcode.org or localhost URL. 

Anything I'm missing here?


src/_static/js/launchcode.js
```diff
@@ -42,9 +42,26 @@ function insertAdmonitionIndicators() {
...
 }
 
+function addTargetToExternalLinks() {
+
+  const links = document.getElementsByTagName("a");
+
+  for (let i = 0; i < links.length; i++) {
+    const link = links[i];
+    const url = link["href"];
+    if (!(url.includes("launchcode.org") || url.includes("localhost"))) {
```
welzie: Can you add `file://`? I don't serve my local site using a web server, I load it directly from the file system.

chrisbay: Sure thing.

chrisbay:
@welzie To my knowledge, and after having conducted a bit of research, I don't think this affects accessibility in any way. There _are_ some general arguments out there about whether the choice of when and whether to use target=_blank. Some argue this should be left to the user. In a general site, I agree with this sentiment. For our educational context though, where we expect people to read a chapter from start to finish, more or less, I think using target=_blank makes sense.

Here are some additional methods to communicate behavior to the user:
- Add a visual indicator, such as an icon, to links opening in a new window (e.g. Wikipedia's arrow box thingy)
- Let the user turn it off by setting a preference in a cookie 
- Don't set target=_blank anywhere, but use an icon to indicate external links to the user can make a more informed decision (a la Wikipedia)

I'm interested in perspectives from @jimflores5 and @gildedgardenia on this.

jimflores5:
I could go for the icon idea, either something like *link* + *icon*, where the link opens in the current tab and the icon opens a new tab, or just having an icon to show that clicking the link will open a new tab.

Would we need our own "arrow-box thingy" if Wikipedia has a copyright? Maybe a rocket-box thingy?  ;-)

Thumbs down on the cookie idea.  I'm pretty sure most users would ignore that option, and including it would add unnecessary time/detail/complexity on our end.

chrisbay:
I merged these changes into `style-changes`, so they will be PRed from there.

PR #32, opened 2019-05-17
Extensions

chrisbay:
Includes 4 local Sphinx extensions. The first two replace the same functionality that was previously applied via JS. Old functionality and styles have been cleaned up.

One question to address is whether these extensions should be added to the `curriculum_module_template` repo. 

## external_links

Adds `target=_blank` and an indicator icon to all external links.

## admonition_icons

Insert icons in front of widely-used admonition titles (warning, tip, note, and a few custom admonitions).

## replit

Creates a new directive, `replit`, which extends the `sourcecode` directive to include a link to a repl.it project. The project is identified by the `:slug:` option.

**Sample RST**

```
.. replit:: js
   :slug: Boolean-Type-Conversion

      console.log(Boolean("true"));
      console.log(Boolean("TRUE"));
      console.log(Boolean(0));
      console.log(Boolean(1));
      console.log(Boolean(''));
      console.log(Boolean('LaunchCode'));
```

**Screenshot**

Screenshot 2019-05-17 16 53 55

## ordered_toctree

Extends the HTML builder to create a builder, `html-a11y`, that uses `ol` elements for toctrees.


jimflores5:
The extensions all seem like reasonable ideas, and they seem to make the book smoother.  

One question I have is, after applying the .. replit:: directive to the new content, how do we want to approach refactoring all of the old examples?

If the new extensions work for LC101 v2, I don't see why we wouldn't want to make them standard and add them to ``curriculum_module_template``.

PR #40, opened 2019-05-24
Assignment 2

welzie:
Assignment 2 write up.

Here is an example solution: https://repl.it/@welzie/scrabble-scorer-v2-solution-practice


src/assignments/scrabble-scorer.rst
```diff
@@ -0,0 +1,157 @@
+Assignment #2: Scrabble Scorer
+==============================
+
+To see how far you skills have come, we are going to give you a mission to update
```
jimflores5: Typo: "your" instead of "you" for the 5th word.

src/assignments/scrabble-scorer.rst
```diff
@@ -0,0 +1,157 @@
...
+that word. Below are the requirements for the new version.
+
+
+Requirements
+------------
+
+#. Create an ``oldScoreKey`` object that contains the previous scoring system.
+#. Create a ``transform`` function which will return a ``newScoreKey`` object.
+#. Create ``scoringAlgorithms`` array to hold three scoring objects, which include the scoring functions.
+#. Create a ``initialPrompt`` function that asks user what score algorithm to use.
```
jimflores5: Typo fixes: "...that asks the user which scoring..."

src/assignments/scrabble-scorer.rst
```diff
@@ -0,0 +1,157 @@
...
+
+   Enter a word to be scored:
+
+
+Detailed Requirements
+---------------------
+
+Transform
+^^^^^^^^^
+Currently our Scrabble Grader uses this data structure when calculating scores
```
jimflores5: Rephrasing suggestion.  How about, "Our current Scrabble Scorer uses the standard rules to calculate the points for a word. The data structure is as follows:"

src/assignments/scrabble-scorer.rst
```diff
@@ -0,0 +1,157 @@
+Assignment #2: Scrabble Scorer
+==============================
+
+To see how far you skills have come, we are going to give you a mission to update
+our *super* important Scrabble Scorer program. What, did you think you were going
+to work on Mars rover code already?
+
+The current Scrabble Scorer program will take in a word and return the Scrabble value for
```
jimflores5: I think we should give some justification for WHY we need to redo the scoring algorithm. Is there a problem with having the points as the keys? Do we want to use different points for certain letters? Will the new algorithm boost efficiency?

If we want to make the assignment sound more realistic, then we should provide a reason why the change needs to be made.

src/assignments/scrabble-scorer.rst
```diff
@@ -0,0 +1,157 @@
...
+   * - Simple Score
+     - Each letter is worth 1 point.
+     - A function with a ``word`` parameter that returns a score.
+   * - Bonus Vowels
+     - Vowels are 3 pts, consonants are 1pt.
+     - A function with ``word`` parameter that returns a score.
+
+.. admonition:: Example
+
+   Example of ``scoringAlgorithms`` object usage.
```
jimflores5: It would be better to give an example from each of the scoring algorithms.  That way the students know if their different functions work properly.

src/assignments/scrabble-scorer.rst
```diff
@@ -0,0 +1,157 @@
...
+#. Create a ``transform`` function which will return a ``newScoreKey`` object.
+#. Create ``scoringAlgorithms`` array to hold three scoring objects, which include the scoring functions.
+#. Create a ``initialPrompt`` function that asks user what score algorithm to use.
+#. Create a ``runProgram`` function to contain the starting point for your running program.
+
+Example Output
+^^^^^^^^^^^^^^
+::
+
+   Welcome to the Scrabble score calculator. Use ctrl+c to quit.
```
jimflores5: Instead of using ctrl-c to stop the program, how about adding 'quit' option to the scoring selection menu?  Maybe option 3 could be 'end' or 'quit'.

src/assignments/scrabble-scorer.rst
```diff
@@ -0,0 +1,157 @@
...
+      5: ['K'],
+      8: ['J', 'X'],
+      10: ['Q', 'Z']
+   };
+
+Write a ``transform`` function that takes in as a parameter the ``oldScoreKey``
+object. ``transform(oldScoreKey)`` will return an object with the letters as keys. The keys
+in the ``newScoreKey`` should be *lowercase* letters.
+
+Example of new key storage
```
jimflores5: Definitely, we need to add a justification for why we want to replace 7 key/value pairs with 26.

src/assignments/scrabble-scorer.rst
```diff
@@ -0,0 +1,157 @@
...
+
+Transform
+^^^^^^^^^
+Currently our Scrabble Grader uses this data structure when calculating scores
+using the standard Scrabble scoring algorithm.
+
+.. sourcecode:: js
+   :linenos:
+
+   const oldScoreKey = {
```
jimflores5: Should we have a "space" option for the blank tiles? ``0: ' '``

src/assignments/scrabble-scorer.rst
welzie: How do you feel about this? "Currently the Scrabble Grader software uses the below data structure for the traditional
Scrabble scoring algorithm. Take a few moments to review how ``oldScoreKey`` relates a
letter to a point value."

src/assignments/scrabble-scorer.rst
welzie: What do you think of this? "The current Scrabble Scorer, is a program that will take in a word and return the point value.
The program needs to be rewritten for two reasons. First the data structure used to store
the letter point values needs to be replaced. Second the program should have three scoring
algorithms and allow the user to pick which algorithm to use. These new features will make
the program more efficient and user friendly."

src/assignments/scrabble-scorer.rst
welzie: I'm not opposed to this. Let's wait and see if some of our QA students also have the same feedback.

src/assignments/scrabble-scorer.rst
welzie: What do you think of this? "The new data structure for storing point values is more efficient. Can you think of why?
With the old data structure, to find the point value for a letter you would need to iterate
over each point value and see if the letter is found in the array of letters that have
that value. In the new data structure has the letters themselves as keys which makes it
much faster to lookup the point value."

src/assignments/scrabble-scorer.rst
welzie: I'll add it as a bonus mission.

src/assignments/scrabble-scorer.rst
welzie: I added examples of all three algorithms scoring the word "JavaScript"

src/assignments/scrabble-scorer.rst
jimflores5: This sounds good, and I like the "new features" approach.  However, there is still a question about WHY the data structure needs to be replaced ("First..."). The second reason might be enough justification on its own.

src/assignments/scrabble-scorer.rst
jimflores5: OK.

src/assignments/scrabble-scorer.rst
jimflores5: I'd prefer to rearrange and reduce your words a bit.  How about, "The current Scrabble Grader uses the data structure shown below. Take a moment to review how ``oldScoreKey`` relates a letter to a point value."

src/assignments/scrabble-scorer.rst
```diff
@@ -0,0 +1,197 @@
...
+The ``newScoreKey`` object will have 26 keys, one for each letter. The value associated to each key
+will be the Scrabble point value.
+
+Example of new key storage
+
+* ``a`` is worth ``1``
+* ``j`` is worth ``8``
+* ``k`` is worth ``10``
+
+The new data structure for storing point values is more efficient. Can you think of why?
```
jimflores5: If we provide the answer in the very next paragraph, then we should remove the rhetorical question, "Can you think why?"

src/assignments/scrabble-scorer.rst
```diff
@@ -0,0 +1,197 @@
...
+With the old data structure, to find the point value for a letter you would need to iterate
```
jimflores5: With the old format, to find the point value for a letter, the program must iterate over each key in ``oldScoreKey`` and check if the letter is inside the array paired with that key.  This "search within a search" is inefficient.

In the new data structure, the letters themselves are keys, so a single search will identify a point value.

src/assignments/scrabble-scorer.rst
welzie: I want to both make them think and give them the answer. Should I say "Before moving on, take a moment to think about WHY the new structure is more efficient."

src/assignments/scrabble-scorer.rst
welzie: Let me know if this comment still applies to the latest version.

src/assignments/scrabble-scorer.rst
welzie: I switched it to no longer ask a question.

src/assignments/scrabble-scorer.rst
jimflores5: "below" needs to come after "data structure".

PR #112, opened 2019-07-23
Appendices updates

gildedgardenia:
Made updates to sourcecode formatting on array and string method appendices.


jimflores5:
Good work, Sally. Thanks for updating the output style.

Next tedious task: switch the **bold** heading to ``inline``.  Currently, the method names appear bold in the TOC and gray in the page headings.  This was a choice we made back when the highlighting was red (which caused potential color perception issues), and we never got around to switching back to `` from ** when we changed to the green scheme.

PR #120, opened 2019-07-30
Terminal commands appendix

CarlyLanglois:
of note, we should check for alternative commands once a windows option is picked


CarlyLanglois:
@gildedgardenia im going to close this since the detailed tutorial is now on a different branch.
You said you want to include the table elsewhere in the chapter no?

PR #121, opened 2019-07-31
add rule for section span child

CarlyLanglois:
adding this rule onto the child span of the section gives the same kind of spacing that you see with the headerlink ref tags. I dont see the difference in the padding pixel height between the two kinds of refs either on firefox or chrome. 

But I do notice that both land here on first click:

Screen Shot 2019-07-31 at 11 42 20 AM

then here on subsequent clicks:

Screen Shot 2019-07-31 at 11 42 34 AM


chrisbay:
I'm seeing less consistent behavior. While sometimes the header is properly positioned, more often than not is is too far down (as in the first screenshot above).

I haven't noticed a difference between first/second click, or between clicking or refreshing the page (at the given header ref). 

I'm not sure how to proceed, but I feel like we're still not a solution.

CarlyLanglois:
closing this since its not as reliable as i'd thought

PR #162, opened 2019-08-22
First draft of CSS Exercises

gildedgardenia:


src/chapters/css/exercises.rst
```diff
@@ -0,0 +1,13 @@
+Exercises: CSS
```
jimflores5: Cool exercises, Sally.  I only found one grammar edit.

src/chapters/css/exercises.rst
```diff
@@ -0,0 +1,13 @@
...
+==============
+
+We have built a website for you to test your CSS knowledge on. Check out the `repl.it <https://repl.it/@launchcode/CSSExercises/>`_!
+
+For the exercises, add the following style rules to ``style.css``:
```
jimflores5: There are multiple tasks for styling elements (``p``, ``h1``, etc.) and one for styling an id, but there are no direct examples for styling a *class*. Students could create a ``centered`` class, for example, and apply that inside the tags, or they could add ``text-align: center;`` to each element in the CSS.

I would vote for adding another ``id`` example and a couple of ``class`` examples in the exercises, especially if they can be structured to reinforce the priority for element/class/id styling.

src/chapters/css/exercises.rst
```diff
@@ -0,0 +1,13 @@
...
+
+1. Change the background color to yellow.
+2. Change all paragraph text color to red.
```
jimflores5: OUCH! Red text on a yellow background. My eyes are hurting just thinking about this. ;-)

src/chapters/css/exercises.rst
```diff
@@ -0,0 +1,13 @@
...
+3. Change all ``h1`` to 36 px font size.
+4. Align all text to the center of the page.
```
jimflores5: What do you think about making this step have multiple parts?  I can think of at least three ways to center all the text on the page.  Perhaps we can ask the students to do something similar?

Example:
4. Align all text to the center of the page.
    a. Now find a *different* way to center the text on the page.
    b. Can you find *another* way to center the text on the page?

src/chapters/css/exercises.rst
```diff
@@ -0,0 +1,13 @@
+Exercises: CSS
+==============
+
+We have built a website for you to test your CSS knowledge on. Check out the `repl.it <https://repl.it/@launchcode/CSSExercises/>`_!
+
+For the exercises, add the following style rules to ``style.css``:
+
+1. Change the background color to yellow.
+2. Change all paragraph text color to red.
+3. Change all ``h1`` to 36 px font size.
```
jimflores5: How do you feel about introducing responsive text sizing as a subset of this?  I remember an exercise in LC101 where we changed the window size to see how the text responded using ``36 px`` vs. something like ``2.5vw``.

src/chapters/css/exercises.rst
```diff
@@ -0,0 +1,13 @@
+Exercises: CSS
+==============
+
+We have built a website for you to test your CSS knowledge on. Check out the `repl.it <https://repl.it/@launchcode/CSSExercises/>`_!
```
jimflores5: Grammar: Remove ending preposition.

How about, 
a) "We built a website for you to test your CSS knowledge."
b) "We created a website where you can practice your CSS styling."
c) "ARRR! Use yer CSS skills ter make yonder website shipshape."
d) Other... Just don't end with a preposition.

src/chapters/css/exercises.rst
```diff
@@ -0,0 +1,13 @@
...
+
+For the exercises, add the following style rules to ``style.css``:
+
+1. Change the background color to yellow.
+2. Change all paragraph text color to red.
+3. Change all ``h1`` to 36 px font size.
+4. Align all text to the center of the page.
+5. Change the font color of elements with the id, ``cool-text``, to blue. 
+
```
jimflores5: I noticed that there is a ``script.js`` file in the repl.it and a ``<script>`` element in the html. Were you going to have the students add some JavaScript to the practice page?

src/chapters/css/exercises.rst
gildedgardenia: Thanks, Jim!

src/chapters/css/exercises.rst
gildedgardenia: Very true! Maybe I will make it green on yellow.

src/chapters/css/exercises.rst
gildedgardenia: We didn't cover this in the chapter.

src/chapters/css/exercises.rst
gildedgardenia: Nope! Repl.it just wants it there and doesn't like it when I remove it.

src/chapters/css/exercises.rst
gildedgardenia: They only learn how to use ``text-align``.

jimflores5: Yes, but you can ``text-align`` each tag, ``text-align`` a class, or encase the tags in one ``div`` element and ``text-align`` that.

src/chapters/css/exercises.rst
jimflores5: Ah. I see.

Could we add some code?  If not, maybe add a disclaimer about what the file would be used for if we wanted it.

src/chapters/css/exercises.rst
gildedgardenia: I think I got some good new exercises! Thanks!

src/chapters/css/exercises.rst
gildedgardenia: Done! Thanks!

src/chapters/css/exercises.rst
gildedgardenia: I add a note to it! Thanks!

src/chapters/css/exercises.rst
gildedgardenia: I added an exercise to use a CSS id to align only the headings to the center.

PR #210, opened 2019-09-18
TypeScript Functions

gildedgardenia:
Especially hoping for insight on what to include and not to include. I tend to want to talk about all of the things with TypeScript!


src/chapters/typescript/functions.rst
```diff
@@ -0,0 +1,115 @@
+Functions in TypeScript
+=======================
+
+When creating functions in TypeScript, we have many of the same options as in JavaScript.
+We can make anonymous functions, give it a different number of parameters and so on.
```
CarlyLanglois: 'give them' rather than than 'give it'

src/chapters/typescript/functions.rst
```diff
@@ -0,0 +1,115 @@
...
+However, when working with functions in TypeScript, we have to keep *type* in mind.
```
CarlyLanglois: maybe *data types* in mind, rather than *type*

src/chapters/typescript/functions.rst
```diff
@@ -0,0 +1,115 @@
...
+
+Let's take a look at a TypeScript function declaration.
+
+.. sourcecode:: typescript
+
+   function myFunction(a: number): number {
+      return a*2;
+   }
+
+Here you can see that we provided the type of the parameters and the type of the value that is returned.
```
CarlyLanglois: I would break this down more since its the first time they are seeing this kind of syntax.
like explicitly, 
"here, the function myFunction has one parameter, ``a``. ``a`` is of type ``number`` and is declared as such. If ``a`` were a string, the function declaration would look like ``myFunction(a: string)``. If myFunction took two parameters, it would be declared as ``myFunction(a: number, b: string)``. myFunction returns a value that is of type ``number``, if it returned an array of numbers, it would be declared as ..."

You get it, just a more fine-grained explanation of what they are looking at in the example could be valuable.

src/chapters/typescript/functions.rst
```diff
@@ -0,0 +1,115 @@
...
+If the function doesn't have a return value, then you use ``void`` for its type.
```
CarlyLanglois: similarly, this warrants an example directive of its own. I don't think the idea of something called ``void`` has been introduced yet.

src/chapters/typescript/functions.rst
```diff
@@ -0,0 +1,115 @@
...
+
+.. note::
+
+   While in many cases, it is possible to determine the return type during compilation, if you know the return type, you should give it.
```
CarlyLanglois: also an example to showcase the syntax if you don't know the return type?

src/chapters/typescript/functions.rst
```diff
@@ -0,0 +1,115 @@
...
+^^^^^^^^^^^^^^^^^^^
+
+For an anonymous function, you still need to provide types for the value returned and the parameters.
+
+.. sourcecode:: typescript
+
+   let myFunction = function(a: number): number { return a*2; };
+
+The above code looks very much like an anonymous function declaration with the types included!
+However, TypeScript has arrow notation and here we use the **fat arrow**, ``=>``.
```
CarlyLanglois: can we get away without discussing fat arrows? i know we've talked about this syntax in other areas of the text and managed to avoid it while still getting crucial topics across. In this case, it looks like it might be necessary but i can't tell?

src/chapters/typescript/functions.rst
```diff
@@ -0,0 +1,115 @@
...
+   let myFunction: (a: number) => number = function(a: number): number { return a*2; };
+
+In TypeScript, the keywords that come after the colon are the type. The ``=>`` operator then indicates the return type of the function.
+
+Optional Parameter
+------------------
+
+In JavaScript, you can declare a function with 5 parameters and only give it 2 when it comes time to use it.
+In TypeScript, that is not the case. When you declare a function with 5 parameters, you have to give the function 5 arguments when calling it.
+However, in TypeScript, you can make parameters optional. This means that when you are calling the function, you can leave off the optional parameter.
```
CarlyLanglois: i think combining this sentence and the one before it makes this concept a bit clearer for me. 

"In TypeScript when you declare a function with 5 parameters, you have to give the function 5 arguments when calling it, unless you make some of those parameters optional. As with JavaScript functions, optional parameters can be omitted."

src/chapters/typescript/functions.rst
```diff
@@ -0,0 +1,115 @@
...
+      1
+      8
+      3
+
+In this example, the ``myFunction`` function has two parameters, ``a`` and ``b``. ``a`` is required and ``b`` is optional. When only one argument is provided to the function, then the value of ``a`` is returned.
+When both arguments are provided, then the sum of the 2 arguments is returned.
+
+Another way to deal with a parameter you need to be optional is to give it a default value.
+
+Let's say that in the example above, we wanted to give ``b`` a default value of 5. That way, if no argument is supplied for ``b``, then the value returned is that of ``a+5``.
```
CarlyLanglois: i think this example can be stronger if the return value when the optional parameter is left off is not the same as the first parameter. does that make sense?

CarlyLanglois: oops! this note would apply to the previous example, not the case where b has a default value

src/chapters/typescript/functions.rst
gildedgardenia: Done! Thanks!

src/chapters/typescript/functions.rst
gildedgardenia: Done! Thanks!

src/chapters/typescript/functions.rst
gildedgardenia: I added some more explanation!

src/chapters/typescript/functions.rst
gildedgardenia: Added an example! Thanks!

src/chapters/typescript/functions.rst
gildedgardenia: Removed this note. Reread some other sources and it seems like it is not worth covering.

src/chapters/typescript/functions.rst
gildedgardenia: Done! Thanks!

src/chapters/typescript/functions.rst
gildedgardenia: I think we can! The anonymous function section is short and sweet, just reminding people to include their types!

src/chapters/typescript/functions.rst
gildedgardenia: Added another number to the sum so that the example will reinforce the fact that the parameter is optional!

src/chapters/typescript/functions.rst
```diff
@@ -0,0 +1,139 @@
...
+   With this in mind, the declaration in TypeScript would look like:
+
+   .. sourcecode:: typescript
+
+      function myFunction(x: number): number {
+         return x*2;
+      }
+
+   Here you can see that we provided the type of the parameter and the type of the value that is returned after the colons.
+   If ``x`` was a string or a boolean, then we would replace ``number`` with the data type of ``x``.
```
CarlyLanglois: if ``x`` were ...

src/chapters/typescript/functions.rst
```diff
@@ -0,0 +1,139 @@
...
+Just as we did above with the named function, we need to make sure that we include the data type for the parameters and the return type of the function.
+
+Optional Parameter
+------------------
+
+In JavaScript, you can declare a function with 5 parameters and only give it 2 when it comes time to use it.
+In TypeScript, that is not the case.
+In TypeScript when you declare a function with 5 parameters, you have to give the function 5 arguments when calling it, unless you make some of those parameters optional.
+As with JavaScript functions, optional parameters can be omitted.
+This means that when you are calling the function, you can leave off the optional parameter(s).
```
CarlyLanglois: I think lines 72 and 73 say the same thing, no? i would keep one and lose the other.

src/chapters/typescript/functions.rst
```diff
@@ -0,0 +1,139 @@
...
+Check Your Understanding
+------------------------
+
+.. admonition:: Question
+
+   What is wrong with this function declaration? NOTE there are 3 things that are wrong here.
+
+   .. sourcecode:: typescript
+
+      let myFunction = function(a:string,b??:number) {return a*b};
```
CarlyLanglois: we should also add this concept check question to the canvas quizzes!

src/chapters/typescript/functions.rst
gildedgardenia: Done! Thanks!

src/chapters/typescript/functions.rst
gildedgardenia: Done! Thanks!

src/chapters/typescript/functions.rst
gildedgardenia: Yes!

PR #237, opened 2019-10-18
Assignment 3 revision

CarlyLanglois:
this revision is meant to bring more attention to the requirements of each class and how they relate to each other. 

most of the actual coding tasks stay the same as the original version, but the starter code has changed and the info returned by receive message is changed

starter code replit and solution will be updated but if you code along while reviewing, take the existing solution code for the command class and its unit tests as the starter


src/assignments/mars-rover.rst
```diff
@@ -185,21 +267,102 @@ several functions outside of ``constructor``.
...
+            {
+               completed: true, 
+               roverStatus: {
+                  mode: 'LOW_POWER', 
+                  generatorWatts: 110, 
+                  position: 98382
+               }
+            }
+         ]
+      }
```
jimflores5: How about condensing the response objects into single lines?  Something like
{ completed: true }, 
{ completed: true, roverStatus: {...} }

Spreading the objects out makes the output less clear (too much whitespace).

src/assignments/mars-rover.rst
```diff
@@ -185,21 +267,102 @@ several functions outside of ``constructor``.
...
+   "responds correctly to status check command"
+
+   a. For the ``STATUS_CHECK`` command, ``receiveMessage(message).results`` 
+      includes a ``roverStatus`` object describing the current state of the 
+      rover object --- ``mode``, ``generatorWatts``, and ``position``. The test 
+      should check each of these for accuracy.
+   b. See the :ref:`Rover Command Types <command-types-table>` table for more
+      details.
+
+Test 11
```
jimflores5: Remember to update the GitLab solution to reflect any changes made to the test descriptions and/or their requirements.

src/assignments/mars-rover.rst
```diff
@@ -185,21 +267,102 @@ several functions outside of ``constructor``.
...
+   
+   a. The test should check the ``completed`` property for accuracy and confirm 
+      that the rover position did not change.
+   b. Use the :ref:`Rover Modes table <rover-modes-table>` for guidance on how
+      to handle move commands in different modes.
+
+Test 13
+   "responds with position for move command".
+
+   a. A move command will update the rover's position with the position value in
```
jimflores5: Slight rephrasing: "A ``MOVE`` command will update the rover's position to the value sent with the command."

src/assignments/mars-rover.rst
```diff
@@ -143,37 +193,69 @@ Message Class
...
+   This test confirms that the ``commands`` property of a new message object
+   contains the data passed in from the ``Message(name, commands)`` call.
 
-      let modeCommand = new Command('MODE_CHANGE', 'LOW_POWER');
-      let moveCommand = new Command('MOVE', 12000);
+   a. Hint: Inside this test, you will have to create a ``commands`` array, fill
+      it with some ``Command`` objects, and pass it into the ``Message``
+      constructor.
+
+.. admonition:: Warning
```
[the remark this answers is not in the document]
CarlyLanglois: i like it here because the equivalent warning at the beginning of this assignment comes right before the message class and I like that pattern

PR #299, opened 2020-11-09
Lightly supported intro

CarlyLanglois:
this intro deserves some more touches and edits as we get more familiar with what the independent learning course (unsupported) looks like


gildedgardenia (commented):
I think this looks great! It remains clear with all of the wording changes. One thing I am wondering is if we should have a dedicated page to the independent learning model. That might come later as we work out the details of the course though.

PR #300, opened 2020-11-10
Renaming Errors from TypeError to err

jamesqquick:
It's confusing to name the error as TypeError in the parameter of the catch. We aren't able to decide which type of error to catch so naming it `TypeError` is extremely confusing for students. It's an arbitrary name of parameter not an actual error type.


CarlyLanglois (approved):
Thanks James this is a good catch

PR #344, opened 2021-07-01
Git studio updates

jimflores5:
The current branch contains the *original* studio page followed by the revised format (split over two pages).

Before we merge, we'll need to remove the original page.


src/chapters/git/studio_rev_1.rst
```diff
@@ -233,11 +249,11 @@ Git the Teamwork Started!
...
 
 You've successfully created a new repository in GitHub and pushed content to
-it. Now it's time to grab a partner and start collaborating on the same repo.
+it. Now it's time for you and your partner to start collaborating on the same
+repo.
 
-You are going to simulate a radio conversation between a shuttle pilot and
-mission control. You and your partner will alternate tasks, so decide who will
-be the **Pilot** and who will be the **Control**.
+One of you decided to play the **Pilot**, while the other chose to play the
```
speudusa, line 255: Line 54 they should have decided, so a reminder to look for their upcoming roles instead?

src/chapters/git/studio_rev_1.rst
```diff
@@ -64,14 +69,25 @@ Step 1: Create a New Local Repository
    part of the commands.
```
speudusa, line 70: I think a role tag would be good here:  **Control and Pilot** and for step 2 (line 173)

src/chapters/git/studio_rev_1.rst
```diff
@@ -64,14 +69,25 @@ Step 1: Create a New Local Repository
...
       $ git init
 
+   .. _important-note:
+
+   .. admonition:: Note
+
+      **IMPORTANT**: To avoid confusion later, it's critical that you and your
+      partner give different names to your repositories.
+
+      For the remainder of this studio, we will refer to the repo as
```
speudusa, line 88: nice!

speudusa (commented):
I like where you broke this up between part 1 and part 2.  I think having the goal to get through part 2 is a lot more realistic.  Merge conflicts as a bonus is great.  

After reading the studio, i wonder if we should discuss push and pull a little more in the reading.  It is there, but it's sparse.  Then we get into the studio and it's another very new thing in a studio full of a lot of very new things.

jimflores5:
> After reading the studio, i wonder if we should discuss push and pull a little more in the reading. It is there, but it's sparse. Then we get into the studio and it's another very new thing in a studio full of a lot of very new things.

This is a good point to consider for a larger follow up.  I'll add an issue to the repo so we can come back to this after we merge in the studio changes.

src/chapters/git/studio_rev_1.rst
jimflores5: Good idea.  I added global role tags at the start of steps 1 and 2.

src/chapters/git/studio_rev_1.rst
jimflores5: Thanks!

src/chapters/git/studio_rev_1.rst
jimflores5: Done!  I reworded this to highlight the role tags and to caution the partners to complete their jobs in the recommended order.

PR #374, opened 2022-03-17
Loops studio updates

gildedgardenia:


src/chapters/loops/studio.rst
```diff
@@ -40,51 +43,65 @@ Part A (Put dinner together)
...
-   `Code it at replt.it <https://repl.it/@launchcode/LoopstudiopartB>`__
+1. Write a ``for`` loop that uses ``mealPrintout()`` to format the printing of each meal in ``mealsFor6``.
+2. Write another ``for`` loop that uses ``mealPrintout()`` to format the printing of each meal in ``mealsForX``.
+3. Initialize the two strings, ``password1`` and ``password2``.
+4. Run your program to see how it works!
 
-Part C (Refinements)
---------------------
+Checking Your Work
+------------------
```
gildedgardenia, line 98: Maybe add a link to Courtney's appendix here

src/chapters/loops/studio.rst
```diff
@@ -40,51 +43,65 @@ Part A (Put dinner together)
...
+
+2. Return ``code``.
+
+Part D: Put it all together!
+----------------------------
+
+Finally, turn your attention to ``runProgram()``.
 
-   `Code it at replt.it <https://repl.it/@launchcode/LoopstudiopartB>`__
+1. Write a ``for`` loop that uses ``mealPrintout()`` to format the printing of each meal in ``mealsFor6``.
```
speudusa, line 92: this might not be intuitive for the students at this level.  Could we offer a hint or suggestion?  Maybe something like "the loop will iterate through the 'mealPrintout()' function.   Make sure that you provide 'mealPrintout()' with directions for what to iterate through and a counter inside its ()."  

I am having a hard time thinking of a hint right now without giving it away.  but if you want to workshop it later, let me know!

src/chapters/loops/studio.rst
```diff
@@ -40,51 +43,65 @@ Part A (Put dinner together)
 
         ['apple', 'banana', 'more kale', 'ice cream', 'chocolate', 'kiwi']
 
-`Code it at replt.it <https://repl.it/@launchcode/LoopstudiopartsAandC>`__
 
-2. Use a ``for`` loop to assemble 6 meals.
+2. Inside of ``mealAssembly()``, write a ``for`` loop to assemble ``numMeals`` meals.
 
-   a. The meals must include one item from each of the source arrays.
+   a. The meals must include one item from each category in the ``pantry`` array.
```
speudusa, line 49: i think a hint about nesting loops might help here.  i remember as a student not knowing that you could do this for the studio.  it made for a very rough studio.

maybe something like this? 
The computer needs to know how many crew members to prepare food for and what ingredients.  Consider creating a nested loop that will create a meal for each crew member and then add it into a larger collection of meals.

speudusa (commented):
I think this turned out great!

src/chapters/loops/studio.rst
speudusa: i think it wouldn't hurt.

gildedgardenia: Got it!

src/chapters/loops/studio.rst
gildedgardenia: That's great! Thank you!

src/chapters/loops/studio.rst
gildedgardenia: thanks!

PR #388, opened 2022-06-25
Update mars-rover.rst

lowrymchristian:
Starter code's line 12 is just the closing brackets, so the final instructions for test 1 were incorrect.


codinglikeagirl42 (approved):
Great catch lowrymchristian, thank you!

PR #396, opened 2022-12-12
added github-actions for CI/CD to github pages

jwoolbright23:


.gitignore
```diff
@@ -7,3 +7,5 @@ pdf/
 .idea/
 __pycache__/
 docs/
+Dockerfile
```
codinglikeagirl42, line 10: Just double checking that this is needed in the .gitignore file.

codinglikeagirl42 (approved):
One comment, otherwise looks good.

This is the migration of travis yml file to github actions.

.gitignore
jwoolbright23: Yep. Mainly because it's on my personal machine. You probably would not have one in your local environment.

